PR #1, opened 2017-03-14
First Dockerfile for headless linux agent

sburnicki:
This is a first Dockerfile to run wptagent headlessly with docker.

Traffic shaping does not work, yet, so I disabled in `entrypoint.sh`, so it's not yet finished but a good start to play around.

I successfully ran a first single-step test with native connectivity on my local machine with it.


docker/linux-headless/browsers.ini
```diff
@@ -0,0 +1,3 @@
+[Chrome]
+exe=/usr/bin/google-chrome
```
pmeenan: I think this might need to be ```exe=/opt/google/chrome/chrome```.  Otherwise terminating will try to kill google-chrome processes instead of chrome.

docker/linux-headless/entrypoint.sh
```diff
@@ -0,0 +1,19 @@
...
+
+if [ -z "$LOCATION" ]; then
+  echo >&2 'LOCATION not set'
+  exit 1
+fi
+
+if [ -z "$NAME" ]; then
+  echo >&2 'NAME not set'
+  exit 1
+fi
```
pmeenan: It's not critical that the name is not set, it falls back to use the hostname.

internal/chrome_desktop.py
```diff
@@ -25,7 +25,8 @@
     '--disable-default-apps',
     '--disable-domain-reliability',
     '--safebrowsing-disable-auto-update',
-    '--disable-background-timer-throttling'
+    '--disable-background-timer-throttling',
+    '--no-sandbox'
```
pmeenan: I'm not comfortable adding --no-sandbox to the code generally.  If it is necessary for Docker then we'll need to make it a config option.

docker/linux-headless/browsers.ini
sburnicki: Didn't get the difference, thanks.

docker/linux-headless/entrypoint.sh
sburnicki: Makes sense.

internal/chrome_desktop.py
sburnicki: Makes sense. I would add a "--dockerized" option, as we also might need to change some other bits to run in docker (maybe traffic shaping).
While `--no-sandbox` isn't a good default choice, it shouldn't be problematic inside docker, as the whole container is sandboxed.

sburnicki:
Thanks for your review. I updated the pull requests according to your comments. I also added `tcpdump`, `nodejs` v.7, `npm` and the lighthouse module via npm.

PR #5, opened 2017-03-17
Add bind install

yoavweiss:
While running the agent saw an error indicating that `rndc` is not installed. Adding its installation to the docs and related script.


pmeenan:
Should be able to ignore it.  It just runs that to flush the dns cache if it does happen to be installed.  I'll see if I can get it to not emit an error.

PR #19, opened 2017-04-20
Support for auto-update when running in container

sburnicki:
These commits introduce auto-updates when running in container.
By default the container will be updated by checking out this repository's master branch and installing `apt` and `npm` updates.
The update behavior can be modified:
- If `UPDATE_POLICY` is anything but `auto`, updates won't be executed at all
- With `GIT_REMOTE` and `GIT_BRANCH` another git repository/branch can be selected to fetch updates from.

What I'm still a little concerned about is pulling directly from the master branch. With the old agent, the specific `wptagent XYZ` commits marked a new agent release. IMHO it would be a good idea to create a "release" branch and regularly merge master into this. The agent could then use the release branch for retrieving updates, so not all development commits are directly used in production.


sburnicki:
Any concerns with this? I'll be happy to make adjustments :)

pmeenan:
Sorry, got sidetracked.  LGTM though I'm also a bit split since it feels weird to be updating inside of the docker container and it not being stable.  I like that it's optional though.

I'll consider a release branch of some kind though I run the public agents on master as well (both the agent and the server) so those are kept as stable as possible.

sburnicki:
Updating inside of the container seems to be weird for me as well. However, we'd need a concept for automatically rebuilding the image when chrome updates are available.

AFAIK, dockerhub currently only rebuilds the image when code changes happen or a linked image is rebuild. So we would get wptagent and ubuntu updates, but not necessarily chrome and lighthouse updates.

The alternative would be a time-based rebuild, e.g. every night. So we could set up an automated REST trigger (see `Build Triggers` under `Build Settings` on dockerhub), or a manual build & push.

If we have regular rebuilds I would disable or remove the container-internal update mechanism and do it with docker.

PR #33, opened 2017-08-25
[don't merge] fix: pass blocked URLs to Lighthouse 

patrickhulce:
should fix https://github.com/WPO-Foundation/webpagetest/issues/913, I didn't see any tests is it mostly a manual setup with a WPT server locally?

blocked on a release with https://github.com/GoogleChrome/lighthouse/pull/3125

### DIFF
--- a/internal/devtools_browser.py
+++ b/internal/devtools_browser.py
@@ -365,6 +365,8 @@ class DevtoolsBrowser(object):
                 command.append('--save-assets')
             if self.options.android or 'mobile' not in self.job or not self.job['mobile']:
                 command.append('--disable-device-emulation')
+            if len(task['block']):
+                command.extend(['--blocked-url-patterns=%s' % url for url in task['block']])
             command.append('"{0}"'.format(self.job['url']))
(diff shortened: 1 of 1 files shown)


pmeenan:
No automated tests.

As far as the command-line option, does lighthouse silently ignore unknown command-line options or will it fail to run on earlier releases?  Synchronizing the code with all of the agents having updated to the latest lighthouse could be challenging and break private instances.

patrickhulce:
It will actually work today except when there's just a single `--blocked-url-pattern` since yargs will treat it as a string instead of an array. We could just append a second empty blocked-url-pattern here to force it as an array in earlier versions or wait until the new version is widespread enough to avoid thorny sync issues? 

I'm not aware of any customers that make this a rush, just wanted to make sure it didn't get lost.

pmeenan:
Closing.  Landed the fix [here](https://github.com/WPO-Foundation/wptagent/commit/e1caf64a8f4ca0f1aa8f1beeb5d60fc61cc11e98).

PR #35, opened 2017-08-28
Get Firefox version from Firefox

soulgalore:
In the current version, the version is "null" in the HAR that makes the HAR not following the specification and cannot be showed in http://www.softwareishard.com/har/viewer/.

Let me know if you wanna do this in another way and I'll update the PR!


pmeenan:
I added code to pull the browser version directly from marionette/webdriver for Firefox and Edge: https://github.com/WPO-Foundation/wptagent/commit/7abc736926571d950854fc238d5c8bfa090b0642

PR #37, opened 2017-08-28
Correct screen size for Mac when recording video fixes #36

### LINKED ISSUE #36: Filmstrip view images has wrong dimension for Firefox/Mac OS X
Are they generated through VisualMetrics, meaning the video is also wrong? Is there a way to keep/get hold of the original video? I can see the path in the log but I guess it is moved to the server?

screen shot 2017-08-28 at 10 44 22 am

The screenshot looks ok:
screen shot 2017-08-28 at 10 45 39 am

soulgalore:


pmeenan:
I'm going to poke around a bit to see if there is a way to get the current scale factor instead of assuming it is 2.  Mine when hooked up to a desktop is set to 1 for example which should still work and get cropped after the fact but it would add a lot of CPU to the video capture.  The main problem is if a 5k display has a scale factor > 2 then it will be cropping again.

Worst case we should be able to get the DPR directly from the browser but it would be nice if we could get it from the OS directly.

pmeenan:
I just pushed a change that gets the display DPR from the browser on OSX and uses that to scale the crop rectangle.  Tried it out on native and 2x and looks to be working: https://github.com/WPO-Foundation/wptagent/commit/86652c21c5362f8b6f42ef4a80cfc8b52b640971

Issue #40, opened 2017-09-04
Firefox on Linux pickup requests to https://tracking-protection.cdn.mozilla.net/...


soulgalore:
<img width="950" alt="screen shot 2017-09-04 at 10 57 35 am" src="https://user-images.githubusercontent.com/540757/30020449-1d267fea-9164-11e7-9790-797f9343db8a.png">

I can not reproduce it on my Mac but I guess you can turn it off with  **privacy.trackingprotection.enabled** https://wiki.mozilla.org/Security/Tracking_protection

I'll do a untested PR with that :)

Best
Peter

pmeenan:
@digitarald Do you have any idea how to kill the requests to tracking-protection.cdn.mozilla.net/except-flashallow-digest256/... on Linux?

We have prefs disabling all of the safebrowsing and tracking protection prefs that I could find but it is [still firing them off](https://www.webpagetest.org/result/170905_FC_743699f299f879e2a955cd03e18363b0/1/details/#waterfall_view_step1).

```
user_pref("browser.safebrowsing.downloads.remote.enabled", false);
user_pref("browser.safebrowsing.enabled", false);
user_pref("browser.safebrowsing.phishing.enabled", false);
user_pref("browser.safebrowsing.malware.enabled", false);
user_pref("browser.safebrowsing.remotelookups", false);
user_pref("browser.safebrowsing.provider.mozilla.updateURL ", "");
user_pref("privacy.trackingprotection.enabled", false);
user_pref("privacy.trackingprotection.pbmode.enabled", false);
user_pref("services.sync.prefs.sync.privacy.trackingprotection.pbmode.enabled", false);
```

I could certainly filter them out of the reports but it looks like they are impacting the actual loading.

soulgalore:
Just FYI we've been using the settings in Geckodriver https://github.com/mozilla/geckodriver/commit/2bfdc3ec8151c427a6a75a6ba3ad203459540495#diff-0448b1e0b4c6692f0a18f4ebd3fc3fb3R4 as a base for Firefox setup but it was a while ago and I guess the trackingprotection is newer.

pmeenan:
Turns out that on Linux it doesn't like loading the prefs file from the profile directory.  Setting the prefs through marionette fixes it.  Fix should be landing soon.

soulgalore:
Ah cool, just verified and it looks good, thanks for the fix!

PR #41, opened 2017-09-04
Disable tracking protection #40

soulgalore:
This is untested on Linux, I don't have an instance up and running, sorry.


pmeenan:
Bit of a nit but would you mind moving it up to where the other privacy settings are set so they are all together?

soulgalore:
Ah, sorry I missed we had others. Fixed it now.

PR #63, opened 2017-10-23
Support full URLs and HTTPS in the wpt_server option

josephscott:
The current version of wpt_server only supports bare host names, like `wpt.example.com`.  It also hard codes all wpt_server values to HTTP.  This change allows wpt_server to provide full URLs, with HTTPS, like `https://wpt.example.com/`.

In the case where a bare host name is still provided, the existing code has been preserved.

My Python is still very basic.  If there is a better approach to this I'm happy to have this changed.

### DIFF
--- a/internal/webpagetest.py
+++ b/internal/webpagetest.py
@@ -199,7 +199,14 @@ class WebPageTest(object):
                     value = parts[1].strip()
                     logging.debug('Setting config option "%s" to "%s"', key, value)
                     if key == 'wpt_server':
-                        self.url = 'http://{0}/work/'.format(value)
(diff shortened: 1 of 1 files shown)


josephscott:
If we can get this change in ( or something like it ), I'd appreciate getting new EC2 Linux agent AMIs generated.

pmeenan:
Change looks good.  That said, wpt_url as a parameter lets you pass the full URL.  I'll merge though because it doesn't hurt anything and can maintain compatibility.

As far as AMI's go, no need to spin new ones.  Before launching the agent the AMI always pulls the latest code from trunk (and runs all apt updates).

Issue #70, opened 2017-12-18
[Proposal] New Docker startup setup


soulgalore:
The current setup use Docker environment variables to pass on parameters to the wptagent with the extra-args to pass on all other parameters:

```
docker run -d \
  -e SERVER_URL="http://my-wpt-server.org/work/" \
  -e LOCATION="docker-location" \
  -e NAME="Docker Test" \
  -e EXTRA_ARGS="--validcertificate true --fps 60"
  --cap-add=NET_ADMIN
  wptagent
```

Instead we could pass on all parameters from the CLI to the agent, meaning it would look like this:

```
docker run -d \
  --cap-add=NET_ADMIN
  wptagent --server http://my-wpt-server.org/work/  \
  --location docker-location  \
  --validcertificate true  \
  --fps 60
```

I like that approach because you then pass on the parameters the same way independent of using Docker. I can make a PR if this is something you want.

Best
Peter

timothyclarke:
I'm currently using this container in a kubernetes cluster passing the variables in. Would this still work in that way ?
The only other way I can think would be to pass a file in and have the container read the file just like an ini file

soulgalore:
I think Kubernetes handles standard parameters passed on.

@timothyclarke do it work for you on Kubernetes? I've heard people having problem setting connectivity (using tc) so I thought running on Kubernetes was out of the question?

timothyclarke:
I had to switch off the traffic shaping when using a docker container. I've got open pull requests for helm charts at https://github.com/kubernetes/charts/pull/3130 and https://github.com/kubernetes/charts/pull/3103.

The feedback I've received boils down to 
 - There needs to be a maintainer.  I'd prefer @pmeenan however I'm happy to maintain until he wants to take over
 - Containers should be tagged ideally more frequently than release (which in many occasions is too old) & latest (which can be broken)
 - They aren't keen on the AMI's. My charts are a hybrid approach where WPT server is a docker container. There are linux WPT agents without traffic shaping in the cluster and if you want more than there are the AMI's for other linux or windows variants

miguelfc:
@soulgalore I am also using kubernetes. The only serious issue I have found is the traffic shaping feature. I have it enabled, but it fails randomly. For example, you could have a three-runs test, and it might fail in one or two of the three (which one is random). Other than that, the platform is fairly stable.

timothyclarke:
I don't suppose you made a note if all the passes and failures were on the same or different nodes in the cluster?
[...]

PR #72, opened 2017-12-18
Same pattern to start WPTAgent with or without Docker

soulgalore:
This is an alternative way to handle input parameters for the container, see #70


pmeenan:
Sorry, just cleaning up PR's.  If you still want to land this could you rebase and re-open?

PR #75, opened 2018-01-11
Changed the netlog json name to a json suffix

yoavweiss:
Just a minor change to save me a click on Chrome over OSX when importing network logs :)


pmeenan:
I'll take care of it slightly differently because there is server logic that also looks at the file name.  I can modify the server code that fetches the netlog to swap the extension to .json for the download.

pmeenan:
Fixed it [here](https://github.com/WPO-Foundation/webpagetest/commit/b0aed6f1f7773ee71ddf31a924b1bc623a80edcb) in the server download logic instead.

PR #99, opened 2018-04-11
[WIP] Calculate hero element metrics

wildlyinaccurate:
This is not quite finished, but I'd appreciate some feedback on the approach I've taken. Feedback on coding style is especially appreciated since I'm not super familiar with Python. Is there an auto formatter I can run the code through? To summarise the change:

- There is a new `--heroelements` option which is a JSON-encoded array of `{name, x, y, width, height}` objects, each representing the position of a "hero" element in the viewport.
- The `calculate_visual_metrics` function has been augmented to calculate and report hero element metrics, if the element positions are provided.
- The `calculate_hero_time` function attempts to determine the time at which each hero element was rendered by creating a mask of the hero element position in the final frame, and finding the first frame that has a minimal difference to this frame.

Things that are left to do:

- More error handling rather than assuming the hero objects are valid.
- Account for frames that are smaller than the viewport.

Questions for @pmeenan:

- `frames_match` takes `crop_region` and `mask_rect` parameters. I couldn't get these to work in the way I expected and I just wanted to get something working, so I did the masking myself in `calculate_hero_time`. Is this correct or should I be able to use `frames_match` without doing the masking up front?
- This creates a load of image files in the video directory. Should I clean them up afterwards? Should I even save the files to disk or just do all the operations in memory (possibly relevant to the point above)?


pmeenan:
Sorry I missed this.  I don't think passing --heroelements as an agent option that is global to all tests makes sense.  Maybe also a flag to enable or disable hero times passed down from the server so it can be enabled globally or on a per-test basis (if there is enough overhead to not justify always collecting it).

My recommendation is to break the logic up into 2 parts:

1 - Have the agent extract the hero element positions (always).
* Have some default logic that looks for the largest image and something related to headings
* Allow for server-defined element ID's on a per-test basis
* Allow for page-defined element ID's to be exposed on the DOM
* Collect the size, position, ID and element type of each of the hero elements
* Collect the viewport size (so the elements can be located relative to the viewport)

2 - Pass the hero time information to visualmetrics.py in a json file
* Let visualmetrics do all of the necessary calculations

That will also make testing and debugging a lot easier since the actual processing can be debugged with a stand-alond video file and hero elements json file.

wildlyinaccurate:
Thanks Pat, sounds good to me! I was operating under the (false) impression that visualmetrics.py was invoked for every test run, hence the `--heroelements` option. A JSON file sounds much better.

I'll ping you when I've made some progress on your suggested approach.

wildlyinaccurate:
Hiya @pmeenan! I've re-worked this PR based on the approach you outlined. A couple of questions before I polish this off:

1. Where is the best place to store the hero times once they're calculated? Should they go in the pageData file or in their own file?
2. Would you prefer that the image masks I create during processing are removed afterwards? I don't think they're totally necessary for debugging as you can recreate them easily enough by running visualmetrics.py.

wildlyinaccurate:
Hey @pmeenan, I'm having trouble figuring out how the heck the visualmetrics.py results are stored. You said the other day that I shouldn't have to manually write the hero element timings to a file, but I'm struggling to find the code that writes the other visual metrics to disk (I can only see them in the pageData file).

pmeenan:
Looks like I'm wrong and you'll have to add an output file for the hero timing.  The current agent passes --progress to visualmetrics.py to have it save out the visual progress and then the visual metrics are calculated server-side using the visual progress.

internal/firefox.py
```diff
@@ -369,6 +369,13 @@ def collect_browser_metrics(self, task):
         page_data = self.run_js_file('page_data.js')
         if page_data is not None:
             task['page_data'].update(page_data)
+        if 'heroElementTimes' in self.job and self.job['heroElementTimes']:
+            logging.debug("Collecting hero element positions")
+            hero_elements = self.run_js_file('hero_elements.js')
```
pmeenan: Doesn't it also need to run the selectors provided by the server as heroElements?  As best as I can tell it only ever collects the defaults specified by hero_elements.js.

wildlyinaccurate: Thanks! Forgot that I was mulling over how to do this. Ended up injecting the custom hero element selectors into a global `window.__wptHeroElements` object -- is that ok with you?

wildlyinaccurate: I can only verify in Firefox & Chrome, but I will change to that approach because I like it better :)

wildlyinaccurate:
I've updated the script to take in custom hero selectors as a function argument, so it's all done in one `execute_js` call.

In terms of testing:

 - I've verified that Opera, Firefox, and Chrome work fine on the Linux agent.
 - I don't have access to any Apple hardware, so can't test Safari.
 - I don't have a "spare" Windows machine but I should have some spare time this weekend to see if I can get the agent running in Docker on Windows.
 - I have plenty of Android devices lying around so will attempt to test on one of those as well.

If you have any test benches already set up, I would appreciate it if you were able to perform some simple testing as well since you know the system boundaries better than I do.

Issue #120, opened 2018-05-07
Timeout for test as well as script steps


zeman:
At the moment the `timeout` parameter is multiplied by each step in a script. For poorly written scripts or scripts with lots of `andWait` and `navigate` commands this can lead to unexpectedly long test timeouts.

Would love to see both an overall `timeout` that provides a hard stop for the test and a `timeout_step` that provides a timeout for each individual step.

PR #126, opened 2018-05-21
Add per-job test_time_limit option

wildlyinaccurate:
I haven't tested this at all but just wanted to see if it was a potential solution to #120. What do you think @pmeenan? Am I grossly oversimplifying a complex problem?

FWIW, I would pair this with the following patch to webpagetest/runtest.php to allow it to be specified either per-job or in settings.ini.

```php
if( $test['test_time_limit'] )
    AddIniLine($testFile, 'test_time_limit', $test['test_time_limit']);
elseif( $settings['test_time_limit'] )
    AddIniLine($testFile, 'test_time_limit', $settings['test_time_limit']);
```

### DIFF
--- a/internal/webpagetest.py
+++ b/internal/webpagetest.py
@@ -589,7 +589,10 @@ class WebPageTest(object):
                         task['width'] = job['width'] + 20
                         task['height'] = job['height'] + 120
                 task['time_limit'] = job['timeout']
-                task['test_time_limit'] = task['time_limit'] * task['script_step_count']
+                if 'test_time_limit' in job:
+                    task['test_time_limit'] = job['test_time_limit']
+                else:
+                    task['test_time_limit'] = task['time_limit'] * task['script_step_count']
                 task['stop_at_onload'] = bool('web10' in job and job['web10'])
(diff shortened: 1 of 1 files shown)


pmeenan:
The server-side part was [done a couple of weeks ago](https://github.com/WPO-Foundation/webpagetest/commit/3a24fa0cdad8541078967442471b5d735a0d13df).  I just stalled and didn't implement the agent-side yet.

For the agent I expect it will be a matter of setting an overall end time in webpagetest.py when the task is started and then in "wait for load" in each of the agents, set the timeout to the lesser value of what it currently calculates or the overall time limit.

wildlyinaccurate:
Oh, I totally missed that! I had hooked into `test_time_limit` because it looked like all of the browsers were already using it as a hard limit in `run_task`:

```py
end_time = monotonic.monotonic() + task['test_time_limit']
while len(task['script']) and task['error'] is None and \
        monotonic.monotonic() < end_time:
    # ...
```

But it sounds like you're describing something a bit more sophisticated. I'll leave it with you :)

Issue #123, opened 2018-05-15
ffmpeg won't record video when viewport size exceeds actual screen resolution


j0weiss:
Hi,

we have windows agents with a small screen resolution running and after updating them to use the new wpt agent we didn't get any visual metrics.
After some digging around we found the following error message:

![error message](https://user-images.githubusercontent.com/25281769/40061881-382035bc-585a-11e8-98e6-71a524d107c4.png)

The problem seems to be that `ffmpeg` or `gdigrab` cannot record the video/filmstrip because the specified resolution `1900x950` exceeds the maximum resolution (`1280x1024`) of the agent itself.

With the old agent we didn't have that problem. And according to the [documentation](https://sites.google.com/a/webpagetest.org/docs/using-webpagetest/scripting#TOC-setViewportSize) the desired behaviour in such cases should be:

> If you get black areas on your screen shots then the viewport is larger than the desktop.

We looked into the code and the class `WebPageTest` contains two variables `self.screen_width` and `self.screen_height`.
The method `on_start_recording(self, task)` in the class `DesktopBrowser` starts the recording of the video and sets the width/height of the video to the width/height provided by the started test.
Maybe one could utilize `self.screen_width` and `self.screen_height` here and use them if the screen resolution is smaller than the viewport size?

j0weiss:
Resolved by pull request https://github.com/WPO-Foundation/wptagent/pull/128

PR #128, opened 2018-05-24
Handle exceeding browser dimensions

j0weiss:
Based on the issue #123 I added two things:
* the agent also sends the resolution when running on macOS and
* `ffmpeg` gets updated width and height based on the actual dimensions of the host i.e. if the given dimensions exceed the resolution of the system the actual dimensions will be used.

### DIFF
--- a/internal/desktop_browser.py
+++ b/internal/desktop_browser.py
@@ -56,6 +56,8 @@ class DesktopBrowser(BaseBrowser):
         self.task = None
         self.cpu_start = None
         self.throttling_cpu = False
+        self.screen_width = None
+        self.screen_height = None
         self.device_pixel_ratio = None
         self.stopping = False
         self.support_path = os.path.join(os.path.abspath(os.path.dirname(__file__)), "support")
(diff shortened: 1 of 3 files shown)


internal/desktop_browser.py
```diff
@@ -365,6 +367,19 @@ def on_start_recording(self, task):
                     self.execute_js(SET_ORANGE)
                     time.sleep(1)
                 task['video_file'] = os.path.join(task['dir'], task['prefix']) + '_video.mp4'
+                if platform.system() == 'Windows':
+                    from win32api import GetSystemMetrics
+                    self.screen_width = GetSystemMetrics(0)
+                    self.screen_height = GetSystemMetrics(1)
+                elif platform.system() == 'Darwin':
+                    from AppKit import NSScreen
```
pmeenan: Can you put a try/except around this just in case it's not installed?

internal/desktop_browser.py
```diff
@@ -56,6 +56,8 @@ def __init__(self, path, options, job):
         self.task = None
         self.cpu_start = None
         self.throttling_cpu = False
+        self.screen_width = None
```
pmeenan: Please set the defaults to 1920/1200 here so None can never happen

internal/desktop_browser.py
```diff
@@ -365,6 +367,19 @@ def on_start_recording(self, task):
...
                 task['video_file'] = os.path.join(task['dir'], task['prefix']) + '_video.mp4'
+                if platform.system() == 'Windows':
+                    from win32api import GetSystemMetrics
+                    self.screen_width = GetSystemMetrics(0)
+                    self.screen_height = GetSystemMetrics(1)
+                elif platform.system() == 'Darwin':
+                    from AppKit import NSScreen
+                    self.screen_width = int(NSScreen.screens()[0].frame().size.width)
+                    self.screen_height = int(NSScreen.screens()[0].frame().size.height)
+                else:
```
pmeenan: No need for the else clause if the defaults are set above.

internal/webpagetest.py
```diff
@@ -81,6 +81,10 @@ def __init__(self, options, workdir):
                 from win32api import GetSystemMetrics
                 self.screen_width = GetSystemMetrics(0)
                 self.screen_height = GetSystemMetrics(1)
+            elif platform.system() == 'Darwin':
```
pmeenan: try/except block here as well please.

internal/desktop_browser.py
```diff
@@ -365,6 +367,19 @@ def on_start_recording(self, task):
...
+                    self.screen_width = 1920
+                    self.screen_height = 1200
+                task['width'] = self.screen_width if task['width'] > self.screen_width else task['width']
```
pmeenan: A bit easier to read if this is changed to min(self.screen_width, task['width']) (and for height).

j0weiss:
I added some changes according to your hints.
Thanks for the feedback.

pmeenan:
Looks good, thanks for the patch.

PR #131, opened 2018-06-01
feat: add support for LH v3

patrickhulce:
I'm not 100% sure of the places that consume the JSON output so if there's other spots happy to update them too. Is the LH score at the top of test result based on the `lighthouse_audits.json` output you create? Might be easiest to do another one of those disconnected agents I can play around with the make sure everything looks good?

Could also just add the `--throttling-method=provided` flag without removing the other one for dual compat with 2.x and 3.x?

NOTE: you can install lighthouse v3 with `npm install -g lighthouse@next` :)

[docs on differences between 2 and 3](https://developers.google.com/web/tools/lighthouse/v3/migration#invocation)

### DIFF
--- a/internal/devtools_browser.py
+++ b/internal/devtools_browser.py
@@ -418,8 +418,7 @@ class DevtoolsBrowser(object):
             html_gzip = os.path.join(task['dir'], 'lighthouse.html.gz')
             time_limit = min(int(task['time_limit']), 80)
             command = ['lighthouse',
-                       '--disable-network-throttling',
-                       '--disable-cpu-throttling',
+                       '--throttling-method', 'provided',
                        '--enable-error-reporting',
                        '--max-wait-for-load', str(int(time_limit * 1000)),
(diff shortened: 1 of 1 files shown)


patrickhulce:
also just to double check the expected format of the `lighthouse_audits.json` is something like the below?

```json
{
  "Performance.first-contentful-paint":2525.0414,
  "BestPractices":0.94,
  "Performance.estimated-input-latency":12.8,
  "Performance.speed-index":12177.3775280233,
  "Accessibility":1,
  "Performance.first-meaningful-paint":2738.04945,
  "Performance.first-cpu-idle":2738.04945,
  "ProgressiveWebApp":0.58,
  "SEO":0.89,
  "Performance":0.7,
  "Performance.interactive":3437.4262
}

pmeenan:
Format is correct.  Basically includes the top-level scores as well as period separated performance metrics.

PR #138, opened 2018-06-20
Added a script for updating the agent when running it as a systemd service

j0weiss:
We are running multiple agents for mobile devices on raspberry pis and wanted a mechanism for keeping them up to date.


pmeenan:
lgtm.

FWIW, I usually run the agents in a screen session that is started by a @reboot cron job and the screen session script does an OS and lighthouse update before running, does an agent update hourly and reboots daily.  The script and setup is done for desktop right now with the [wptagent-install script](https://github.com/WPO-Foundation/wptagent-install) and I've been meaning to create a Pi variation of it that does all of the pi-specific setup for mobile as well.

PR #158, opened 2018-07-19
Run visualmetrics.py with verbose logging (-vvvv) when debug=1

wildlyinaccurate:
I find it useful to run visualmetrics.py with -vvvv when I'm debugging. It would be nice if this was done automatically based on the debug option.

The one thing I'm not sure of is whether it's possible to get this output from the WPT UI? I primarily test in a Docker container, so I see all the console output. It would be nice to be able to see the output somehow when running tests on webpagetest.org.

### DIFF
--- a/internal/android_browser.py
+++ b/internal/android_browser.py
@@ -184,6 +184,8 @@ class AndroidBrowser(BaseBrowser):
                         '{0:d}'.format(self.job['imageQuality']),
                         '--viewport', '--maxframes', '50', '--histogram', histograms,
                         '--progress', progress_file]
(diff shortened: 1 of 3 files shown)


pmeenan:
Adding it to the debug log that WPT reports back is possible and is [done for lighthouse](https://github.com/WPO-Foundation/wptagent/blob/master/internal/devtools_browser.py#L439) but it requires plumbing through the stdout/stderr pipes from the process.

pmeenan:
The other option is to call into the python directly instead of through a shell script, then it will inherit the logging of wptagent and be included in the debug log automatically.

Issue #169, opened 2018-08-21
Run custom "hero" elements standalone without default "heroes"


soulgalore:
Hey @pmeenan @wildlyinaccurate I want to suggest a change in how the custom "hero" elements work. At the moment the flow is like this:
Loop through all elements on the page, search for img, h1/h2, background images and attribute element timing. Then use custom selectors to get specific elements.

I want to change this so that if the user add custom selectors only those elements are used. That way we will get rid of the loop of all elements (for https://en.wikipedia.org/wiki/Barack_Obama that is 16000+ elements). On the machine that I tested on that loop adds quite much overhead in total test time. 

And I think if the user chooses to configure custom elements she/he can choose exactly the elements that is needed and there's no need for the automatic catch of metrics?

I'll do a PR and you can discuss if this is fine for you :)

Best
Peter

wildlyinaccurate:
I'm surprised to hear that the loop adds overhead in your tests. Do you mind me asking what your test setup is? I mostly tested this on my Raspberry Pi 2, and the hero element overhead was about 6 seconds including the Python code that does the frame-by-frame image analysis. On our EC2 WPT agents (c4.large) the overhead is more like 2 seconds.

The only problem I have with #170 is that it would make it impossible to have the default elements _and_ custom elements. The default elements are "special" in that they find the largest of each heading, image, and background image, so it's not enough to just have add a custom element with "h1" as the selector.

Maybe an extra boolean attribute to disable the default hero elements would be better? Although I'm still very curious to know what it is about your test setup that causes this code to add so much overhead!

pmeenan:
If the JS is taking a long time to run (maybe on a mobile device or when CPU throttling is active) I'd love to see if there is anything we could do to optimize the search so maybe it didn't have to look at all 16000+ elements (maybe with a query selector of some kind).

wildlyinaccurate:
The background image detection is the only reason to enumerate every node in the document, since there's no way (that I know of..?) to select only elements with a background image. I ran a few benchmarks on my RPi 2. The average run time for the JS on https://en.wikipedia.org/wiki/Barack_Obama is 380ms. Without the background image detection it's 10ms.

The real issue is calling `getBoundingClientRect()` on every node. It's much quicker to use `IntersectionObserver` but I'm not sure how to deal with it being async. Is there a synchronous way to check whether an element is in the viewport that is quicker than `getBoundingClientRect()`?

pmeenan:
[...]

PR #170, opened 2018-08-21
If custom elements are configured, only those elements are collected

soulgalore:
See https://github.com/WPO-Foundation/wptagent/issues/169


pmeenan:
Sorry, cleaning up PR's.  I'll ping the issue to see if it is still a problem and see how we want it to be addressed.

PR #174, opened 2018-08-27
Increase hero frame match fuzz to 5%

wildlyinaccurate:
I've been slowly working through some anomalous hero element times and most of them appear to be "fuzziness" issues. For example, the background image behind the H1 on this page is _nearly_ the same colour as the body background colour.
 
hero-fuzz

There's nothing scientific about the way I picked the fuzz factor. For the frames above, 3% fuzz factor is enough to match the frames. @pmeenan I'm curious whether you have any opinions on what a good fuzz factor is? I'm being pretty conservative since `frames_match` will be working with PNGs so there should be no compression artefacts to "fuzz out".


pmeenan:
Sorry, dropped off my radar - LGTM.  FWIW, I use 10% for start render and the last frame so 5% is still pretty conservative.

wildlyinaccurate:
Good to know - thanks!

PR #178, opened 2018-09-07
Pass custom user agent through to Lighthouse via --chrome-flags

wildlyinaccurate:
This solves the (somewhat esoteric) problem of wanting Lighthouse to use custom UA strings rather than the default Chrome string.

Kind of a pain that this is only possible when device emulation is disabled, though. Maybe I'm better off seeing if the Lighthouse team would be open to a `--user-agent` or similar flag that could set the UA via puppeteer?

### DIFF
--- a/internal/devtools_browser.py
+++ b/internal/devtools_browser.py
@@ -477,6 +477,8 @@ class DevtoolsBrowser(object):
                 command.append('--save-assets')
             if self.options.android or 'mobile' not in self.job or not self.job['mobile']:
                 command.append('--disable-device-emulation')
+                if 'user_agent_string' in self.job:
+                    command.append('--chrome-flags="--user-agent=\'{0}\'"'.format(self.job['user_agent_string']))
             if len(task['block']):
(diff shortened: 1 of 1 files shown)


pmeenan:
Can you do some sanitizing to make sure there are no quotes in the UA string?  Otherwise it becomes an avenue for attack on the agents and will allow for running arbitrary code if someone abuses it.

wildlyinaccurate:
Good call - thanks. Does this regex look okay to you?

pmeenan:
LGTM, thanks.

pmeenan:
Actually, thinking about it (because of the linked issue), this shouldn't work.  wptagent launches the browser and passes the dev tools interface to Lighthouse so the flag would need to be added in Chrome mobile/desktop when it launches the browser (with a check that it is running a lighthouse test).

wildlyinaccurate:
Oh, that's a bummer. Apologies for the useless patch! I wasn't able to perform an E2E test for this changeset; I only confirmed that the flag was being set.

Issue #160, opened 2018-07-24
OCSP requests on Firefox aren't ignored for TTFB


sburnicki:
When testing a page with Firefox, the OCSP request accounts for the TTFB:
http://www.webpagetest.org/result/180724_TG_377229a501858985c049f7612ffd94ea/1/details/#waterfall_view_step1

It should be ignored, so the TTFB of the real HTML requests is reported.

stephendonner:
@sburnicki @bwollmer (and @pmeenan): should this be now (marked) fixed?

pmeenan:
Yep, looks to be handled.

PR #180, opened 2018-09-19
added check to firefox to ignore ocsp/pkix-crl for ttfb

bwollmer:
This should solve the following issue:
https://github.com/WPO-Foundation/wptagent/issues/160

### DIFF
--- a/internal/firefox.py
+++ b/internal/firefox.py
@@ -1044,7 +1044,9 @@ class Firefox(DesktopBrowser):
             elif request['responseCode'] > -1:
                 page['responses_other'] += 1
             if main_request is None and \
-                    (request['responseCode'] == 200 or request['responseCode'] == 304):
+                    (request['responseCode'] == 200 or request['responseCode'] == 304) and \
+                     request['contentType'] != 'application/ocsp-response' and \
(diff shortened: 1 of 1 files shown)


internal/firefox.py
```diff
@@ -1044,7 +1044,9 @@ def calculate_page_stats(self, requests):
             elif request['responseCode'] > -1:
                 page['responses_other'] += 1
             if main_request is None and \
-                    (request['responseCode'] == 200 or request['responseCode'] == 304):
+                    (request['responseCode'] == 200 or request['responseCode'] == 304) and \
+                     request['contentType'] != 'application/ocsp-response' and \
+                     request['contentType'] != 'application/pkix-crl':
```
pmeenan: Can you also add a check to make sure "'contentType' in request" before checking the values?  Otherwise it will throw an exception for responses without a contentType.  It doesn't look like the code above has a default value for it when it's not present so it just won't be populated.

bwollmer: You are right, I forgot about that. I added a check if the key doesn't exist. The version before my PR accepted a request without contentType as an main request, so I assume that this is the desired behavior. I changed my code to match the old behavior. Now a request without contentType or a request with the right contentType can be the main request.

PR #183, opened 2018-09-25
fix edge cookies

PhilAlbrecht:
Added a day to expiration date of the edge cookies, in order for edge cookies to be applied again.


internal/microsoft_edge.py
```diff
@@ -68,8 +68,20 @@ def prepare(self, job, task):
         self.bodies_path = os.path.join(task['dir'], 'bodies')
         if not os.path.isdir(self.bodies_path):
             os.makedirs(self.bodies_path)
-        if not task['cached']:
-            self.clear_cache()
+        try:
+            import _winreg
+            keyval = r"SOFTWARE\Classes\Local Settings\Software\Microsoft\Windows\CurrentVersion\AppContainer\Storage\microsoft.microsoftedge_8wekyb3d8bbwe\MicrosoftEdge\Privacy"
+            if not os.path.exists("keyval"):
```
pmeenan: AFAIK, os.path.exists checks the filesystem (relative to the cwd), not the registry and this will always be false.

internal/microsoft_edge.py
```diff
@@ -68,8 +68,20 @@ def prepare(self, job, task):
...
+                key = _winreg.CreateKey(_winreg.HKEY_CURRENT_USER, keyval)
+            Registrykey = _winreg.OpenKey(_winreg.HKEY_CURRENT_USER, keyval, 0, _winreg.KEY_WRITE)
```
pmeenan: The CreateKey should be replaced with CreateKeyEx with KEY_WRITE and OpenKey can be skipped since CreateKeyEx will open the key if it already exists.

internal/microsoft_edge.py
```diff
@@ -70,16 +72,15 @@ def prepare(self, job, task):
             os.makedirs(self.bodies_path)
         try:
             import _winreg
-            keyval = r"SOFTWARE\Classes\Local Settings\Software\Microsoft\Windows\CurrentVersion\AppContainer\Storage\microsoft.microsoftedge_8wekyb3d8bbwe\MicrosoftEdge\Privacy"
-            if not os.path.exists("keyval"):
-                key = _winreg.CreateKey(_winreg.HKEY_CURRENT_USER, keyval)
-            Registrykey = _winreg.OpenKey(_winreg.HKEY_CURRENT_USER, keyval, 0, _winreg.KEY_WRITE)
+            registry_key = _winreg.CreateKeyEx(_winreg.HKEY_CURRENT_USER, self.edge_registry_path, 0, _winreg.KEY_READ)
```
PhilAlbrecht: since i cant read values which have a KEY_WRITE access, i thought the cleaner approach would be to first set read and than switch to write rights.

pmeenan: You can or the access rights with each other: ```_winreg.KEY_READ | _winreg.KEY_WRITE``` so the same key can both read and write.

internal/microsoft_edge.py
```diff
@@ -199,6 +198,10 @@ def stop(self, job, task):
                     os.remove(self.wpt_etw_done)
                 except Exception:
                     pass
+        import _winreg
```
pmeenan: Any chance of this throwing an exception?  The setup is inside of a try/except block.

PhilAlbrecht: I guess better safe then sorry and put it into on. Oriented myself on the win32inet block on the edge.py code.

internal/microsoft_edge.py
```diff
@@ -68,8 +70,18 @@ def prepare(self, job, task):
...
         if not os.path.isdir(self.bodies_path):
             os.makedirs(self.bodies_path)
-        if not task['cached']:
-            self.clear_cache()
+        try:
+            import _winreg
+            registry_key = _winreg.CreateKeyEx(_winreg.HKEY_CURRENT_USER, self.edge_registry_path, 0, _winreg.KEY_READ | _winreg.KEY_WRITE)
+            self.edge_registry_key_value = _winreg.QueryValueEx(registry_key, "ClearBrowsingHistoryOnExit")[0]
+            if task['cached'] or job['fvonly']:
+                _winreg.SetValueEx(registry_key, "ClearBrowsingHistoryOnExit", 0, _winreg.REG_SZ, "1")
```
pmeenan: Everything I can find online says it is a DWORD value key (which is normal for flags).  Unless you can find documentation that shows it is a string key, all of these should be writing it as a DWORD.

PhilAlbrecht:
Yes, we tested it on a up-to-date Windows10 installation. Works for First View only and Repeated View. Also resets the Edge setting to the setting it was before the test. I apologize for the inconvenience. I test the code directly on the test machine and commit it through my work computer, typo must have slipped in when i quit the remote connection.

internal/microsoft_edge.py
PhilAlbrecht: Will change and test it on monday, when i have access to the windows10 system.

pmeenan:
Great, thanks.  Was just getting nervous.

PhilAlbrecht:
Changed the type and tested repeated view and first view only on the latest Windows 10 version. Running the agent with the batch script you recommended on the installation guide. Everything worked just fine.

pmeenan:
Sorry to keep this dragging but it looks like the last change also changed the line endings.  Any chance you can switch them back?

PhilAlbrecht:
Ok I dunno how that happened, but I changed the line breaks. Sorry for that.

internal/microsoft_edge.py
```diff
@@ -68,8 +70,18 @@ def prepare(self, job, task):
         self.bodies_path = os.path.join(task['dir'], 'bodies')
         if not os.path.isdir(self.bodies_path):
             os.makedirs(self.bodies_path)
-        if not task['cached']:
-            self.clear_cache()
```
pmeenan: Sorry to keep dragging this out.  Can you also leave the explicit clear in place as both a backup and to catch anything that the reg key may not clear?

pmeenan:
Perfect, thank you (and thanks for putting up with the nits)

PR #188, opened 2018-10-05
Fix flake8 issues in wptagent.py

stephendonner:
@pmeenan @davehunt @whimboo r?

### DIFF
--- a/wptagent.py
+++ b/wptagent.py
@@ -16,6 +16,7 @@ import sys
 import time
 import traceback
 
+
 class WPTAgent(object):
     """Main agent workflow"""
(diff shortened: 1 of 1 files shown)


stephendonner:
@davehunt go for it; @pmeenan has ultimate say, of course, but it would be great for code readability, which I'm doing a lot of here, lately ;-)

davehunt:
@stephendonner great, I'd start by reading over the links in my previous comment, and also looking at where we did this for another project. One example would be https://github.com/mozilla/FoxPuppet/commit/d869b5691323f385dfff0d351bb792bc73d57cba

Issue #189, opened 2018-10-08
Firefox multistep: Request in wrong step


bwollmer:
If you run a multi step script in firefox, it seems like some requests of the previous step will be in the following step. Here is an example:
https://www.webpagetest.org/result/181008_G3_d3a8c4ca14013a4a17beb2ea4dd45b57/1/details/ 
On the screenshot below or step PL:::OTTO_Produktliste_geschirrspueler the first request should't be there. If you inspect the request, you see a negative start time, so I assume that the request belongs to the previous step. The same for the next 2 requests.
![negative_start](https://user-images.githubusercontent.com/12231522/46603824-8e1e9500-caf4-11e8-9954-79e69ba69e3c.png)

I tried to debug this behavior, but I'm kind of stuck and would appreciate some help. 
In firefox.py:415 the negative time is calculated, because the task start time is greater than the event time.
But I don't know why the message even arrived in this method. Sure, we got it from the message server, but if the problem would be in the message server, this behavior should also appear in other browsers. So I would assume that the problem has to be in firefox.py.
Do you have any idea, where I could look further or why this message arrived in _process_message_?

bwollmer:
Thanks for your feedback! The log seems fine, there is already a line to filter old data out.
We found out that the extensions still reports data, while the current data is processed. In that case the message will be in the queue of the message server, until the next steps starts, which then polls the queue and will receive the old data.
Edge already implemented the logic to empty the queue before each step, I tried to apply the same logic to firefox. This reduced the problem to fewer messages from old steps, but this is just "luck" because firefox still runs in the background and we just minimized the time frame for the concurrent push to the queue.
To get a deterministic result I implemented a filter in firefox.py:process_messages, to filter messages with a negative timestamp. While this solved the problem of old requests in steps, we saw that there is another problem, with and without this fix.

On the first screenshot I posted, you can see that 3 css requests are before the 7. requests. But the 7. requests is the main request and the referrer of the css files, so it"s not possible that the 3 requests happened before that one. We could track this issue down to the merging of log and extension data, but at this point we haven't figured out the exact problem. Below are two screenshots. In both my fix with the old requests is not implemented.  In the first one I disabled all log data and in the second all extension data. In both cases the order seems right. But if both are merged, the order is wrong, like in the first screenshot.

Could you give us some insight, why we have extension and log data and why they are merged like this? The code reads like the log always tells the truth and will overwrite extension data, so do we assume that the extension data is 'wrong'?

[...]

PR #201, opened 2018-10-23
Firefox Multistep: Wrong timings and request order

bwollmer:
This should fix https://github.com/WPO-Foundation/wptagent/issues/189

The firefox code will now only accept extension messages, which where send after the step was started. This should prevent requests, which were started while calculating the current step to appear in the following step.

Additionally all steps after the first one would't start at 0 (see screeshot below), this was fixed by calling adjust_timings for every step, instead of just the first one. This also fixed the wrong order of requests, mentioned in the issue above.

These three bugs where kind of related(see https://github.com/WPO-Foundation/wptagent/issues/189).
ff_offset

### DIFF
--- a/internal/firefox.py
+++ b/internal/firefox.py
@@ -414,10 +414,11 @@ class Firefox(DesktopBrowser):
                     elapsed = event_time - self.task['start_time']
                     message['body']['timeStamp'] = elapsed.total_seconds()
                 cat, msg = message['path'].split('.', 1)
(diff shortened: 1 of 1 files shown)


pmeenan:
Sorry about that.  Merged the change.  My only concern with the adjustment of the timings for steps > 1 is that actions that include a bunch of cached requests before the first network request on an action may be reported incorrectly but that's less of an issue than reporting everything incorrectly.

Issue #203, opened 2018-10-24
CDN Matching can miss identify layered CDNs 


Friss:
The current CDN check will exit early once it finds a match this can be problematic if you layer CDN's. Such as Cloudflare -> Cloudfront -> S3. The script will always match `Amazon Cloudfront` first because its ordered alphabetically. 

https://github.com/WPO-Foundation/wptagent/blob/71f1fd511d360ae8a1d46afd7212214cb7eee5ed/internal/optimization_checks.py#L639-L654

Example (Has both cloudflare & Cloudfront headers)
```
zfriss@BOSMM2TX2V7HTD8 wptagent (master) $ curl -I https://static.hsappstatic.net/StyleGuideUI/static-3.109/img/sprocket/favicon.ico
HTTP/2 200
date: Wed, 24 Oct 2018 02:56:35 GMT
content-type: image/vnd.microsoft.icon
content-length: 15086
set-cookie: __cfduid=d85efe6b6ffb83804dd4a352b8dd5841c1540349795; expires=Thu, 24-Oct-19 02:56:35 GMT; path=/; domain=.hsappstatic.net; HttpOnly
x-amz-replication-status: COMPLETED
last-modified: Thu, 04 Oct 2018 18:21:05 GMT
etag: "85dd40a78c8d902fe31b3519d9f8255b"
x-amz-version-id: cgXdGSWK7SQkfa8N5UULC_9bGccrMKkj
x-cache: Miss from cloudfront
via: 1.1 4ee3d5920fafcf4bca394fd489654c8c.cloudfront.net (CloudFront)
x-amz-cf-id: KMIOV7GpVp3YsEhx43HWT7reRRCOTjzoAkclO1_1EfdVImaGrmzeDw==
cf-cache-status: HIT
expires: Thu, 24 Oct 2019 02:56:35 GMT
cache-control: public, max-age=31536000
accept-ranges: bytes
strict-transport-security: max-age=31536000; includeSubDomains; preload
expect-ct: max-age=604800, report-uri="https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct"
vary: Origin,Accept-Encoding,Access-Control-Request-Headers,Access-Control-Request-Method
server: cloudflare
cf-ray: 46e92bcc3d055687-IAD
```

I'm not sure what the best fix here would be. We could change the script to find all matches and join them together. In which case the script would return `Amazon Cloudfront and Cloudflare`.

[...]

PR #207, opened 2018-10-29
Check all CDN headers for layered CDNs

Friss:
Resolves https://github.com/WPO-Foundation/wptagent/issues/203

Since headers are the only thing that could match against 2 CDNs I just added this logic here and not to the cname checking. 

Before:
screen shot 2018-10-29 at 2 38 40 pm

After:
screen shot 2018-10-29 at 2 38 47 pm


pmeenan:
Thanks for the PR, looks great.

Friss:
Thanks for merging 🥇

PR #219, opened 2018-11-20
Firefox - sets most log levels to 1, from 5; x-ref Bugzilla bug # 1452535

stephendonner:
@davehunt @pmeenan @soulgalore this is, as the title notes, the corresponding WebPageTest-side changes to some Firefox log levels in the client (https://bugzilla.mozilla.org/show_bug.cgi?id=1452535), to address runtime/startup overhead.

r?

### DIFF
--- a/internal/firefox.py
+++ b/internal/firefox.py
@@ -54,8 +54,8 @@ class Firefox(DesktopBrowser):
         self.requests = {}
         self.main_request_headers = None
         os.environ["MOZ_LOG_FILE"] = self.moz_log
-        os.environ["MOZ_LOG"] = 'timestamp,sync,nsHttp:5,nsSocketTransport:5'\
-                                'nsHostResolver:5,pipnss:5'
+        os.environ["MOZ_LOG"] = 'timestamp,sync,nsHttp:1,nsSocketTransport:1'\
+                                'nsHostResolver:1,pipnss:5'
         DesktopBrowser.prepare(self, job, task)
         profile_template = os.path.join(os.path.abspath(os.path.dirname(__file__)),
                                         'support', 'Firefox', 'profile')


internal/firefox.py
```diff
@@ -54,8 +54,8 @@ def prepare(self, job, task):
         self.requests = {}
         self.main_request_headers = None
         os.environ["MOZ_LOG_FILE"] = self.moz_log
-        os.environ["MOZ_LOG"] = 'timestamp,sync,nsHttp:5,nsSocketTransport:5'\
-                                'nsHostResolver:5,pipnss:5'
+        os.environ["MOZ_LOG"] = 'timestamp,sync,nsHttp:1,nsSocketTransport:1'\
+                                'nsHostResolver:1,pipnss:5'
```
davehunt: We'll need to make this conditional based on the version of Firefox as the relevant logs will only appear in this level in the very latest Firefox Nightly builds (v65).

pmeenan: +1.  We may not need to track the actual version number because we need the env to be set before launching.  My vote would be in the [browser detection](https://github.com/WPO-Foundation/wptagent/blob/master/wptagent.py#L480), add a "log_level" to the browser options and set it to 1 for Nightly. Once it rolls to beta then also add it to beta - and then finally to stable.

Then in here, default log_level to 5, override it from the options if present and then format the string with the appropriate log level.

Bonus points if someone validated that all of the necessary log events in each of the relevant categories for WPT have been set correctly on the Firefox side because it may be hard to notice if some events are missing. Worst-case we can see how the waterfalls look for Nightly but that's a bit black-box.

pmeenan:
I just [pushed a change](https://github.com/WPO-Foundation/wptagent/commit/eb40e0281773eeed56069b4b3830d79163496085) that sets it selectively for Nightly.

davehunt:
Thanks @pmeenan!

stephendonner:
Wow, yes - thanks hugely, Pat!

PR #223, opened 2018-12-05
Switch to using multiprocesssing queue in message server

vibaldem:


pmeenan:
Looks good. Eventually it would probably make sense to move to a pipe instead of queue since there is only one reader and one writer but it's not critical.

Issue #229, opened 2019-01-08
Calculate and properly name domContentFlushed metric (Firefox-specific/internal) from timeToDOMContentFlushed


stephendonner:
Now that I've figured out/coordinated with our core Firefox dev + perf-testing team on the metric's calculation, I can finally correct the unwieldy ```timeToDOMContentFlushed``` metric, essentially doing:

```
var perfData = window.performance;
var domContentFlushed = perfData.timing.timeToDOMContentFlushed - perfData.timing.fetchStart;
return domContentFlushed;
```

In here:  https://github.com/WPO-Foundation/wptagent/blob/758685caff1cc54067f931ff6a00d77db2846422/internal/js/page_data.js

which will return a duration in milliseconds, accounting for the DOM-content flushing time

/cc @davehunt

stephendonner:
Fixed in https://github.com/WPO-Foundation/wptagent/pull/230.

PR #230, opened 2019-01-08
[WIP] Add new, correct domContentFlushed (Fx-only) metric; remove raw timeT…

stephendonner:
…oDOMContentFlushed

Attempts to address https://github.com/WPO-Foundation/wptagent/issues/229

This is **untested** in ```WebPageTest```, but the Custom Metrics JS snippet has been running in our private instance for a while now.

This is more of a working state for today's work, so I don't lose it (should work, but doesn't look pretty, nor in prevailing style, yet).

/cc @davehunt

### DIFF
--- a/internal/js/page_data.js
+++ b/internal/js/page_data.js
@@ -24,11 +24,12 @@ addTime("domComplete");
 addTime("loadEventStart");
 addTime("loadEventEnd");
 addTime("timeToFirstInteractive");
-pageData["domContentFlushed"];
-var perfData = window.performance;
-var domContentFlushed = perfData.timing.timeToDOMContentFlushed - perfData.timing.fetchStart;
-return domContentFlushed;
-
+try {
+    if (window.performance.timing['timeToDOMContentFlushed']) {
+        pageData["domContentFlushed"] = window.performance.timing.timeToDOMContentFlushed - window.performance.timing.fetchStart;
+    }
+} catch(e) {
+}
 pageData["firstPaint"] = 0;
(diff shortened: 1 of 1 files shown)


stephendonner:
xref: https://github.com/mozilla/wpt-api/issues/97 and https://github.com/WPO-Foundation/wptagent/pull/205#issuecomment-433208576, particularly

internal/js/page_data.js
```diff
@@ -20,11 +20,15 @@ addTime("domInteractive");
 addTime("domContentLoadedEventStart");
 addTime("domContentLoadedEventEnd");
 addTime("timeToContentfulPaint");
-addTime("timeToDOMContentFlushed");
 addTime("domComplete");
 addTime("loadEventStart");
 addTime("loadEventEnd");
 addTime("timeToFirstInteractive");
+pageData["domContentFlushed"];
```
pmeenan: I'm assuming this is just holding the code from the custom metric because it won't actually work as-written.  Looks like something like this should work:
```JavaScript
try {
    if (window.performance.timing['timeToDOMContentFlushed']) {
        pageData["domContentFlushed"] = window.performance.timing.timeToDOMContentFlushed - window.performance.timing.fetchStart;
    }
} catch(e) {
}

stephendonner: Thanks; I didn't honestly expect it to work either, and my JS is beyond rusty (sorry, Brendan!)

stephendonner: Fixed.

PR #235, opened 2019-01-23
Use pip list --format=freeze for compatibility across pip versions

wildlyinaccurate:
Our EC2 agents just started failing a few hours ago. I think it's due to a pip upgrade where the default format for `pip list` is now "columns", which wptagent.py does not parse correctly. I think wptagent.py knows how to parse the "legacy" format, but this doesn't seem to be available any more in pip 19.0, so I picked the "freeze" format which seems to be available in both 9.0 and 19.0 (the only two versions I have to test with).

Before merging this, would you be able to sanity check that this is a good change? I don't know a whole lot about the pip release cycle or the differences between versions.

### DIFF
--- a/wptagent.py
+++ b/wptagent.py
@@ -696,9 +696,9 @@ def upgrade_pip_modules():
         subprocess.call([sys.executable, '-m', 'pip', 'install', '--upgrade', 'pip'])
         run_elevated(sys.executable, '-m pip install --upgrade pip')
         out = subprocess.check_output([sys.executable, '-m', 'pip', 'list',
-                                       '--outdated'])
+                                       '--outdated', '--format', 'freeze'])
         for line in out.splitlines():
-            separator = line.find(' ')
+            separator = line.find('==')
             if separator > 0:
                 package = line[:separator]
                 subprocess.call([sys.executable, '-m', 'pip', 'install',


wildlyinaccurate:
Hmm, this might be a red herring. The output of `pip list --outdated` on our agents is 

```
Package      Version Latest Type 
------------ ------- ------ -----
cryptography 2.4.2   2.5    wheel
dnspython    1.15.0  1.16.0 wheel
entrypoints  0.2.3   0.3    wheel
idna         2.7     2.8    wheel
keyring      16.1.1  17.1.1 wheel
keyrings.alt 3.1     3.1.1  wheel
mozlog       3.9     3.10   wheel
Pillow       5.3.0   5.4.1  wheel
pygobject    3.26.1  3.30.4 sdist
pyxdg        0.25    0.26   wheel
requests     2.20.1  2.21.0 wheel
setuptools   40.6.2  40.6.3 wheel
six          1.11.0  1.12.0 wheel
```

So the agent tries to run `pip install Package` and `pip install ------------`, which both fail. But where it's actually getting stuck is on `pip install cryptography`, at the point when it's uninstalling the existing package:

```
Collecting cryptography
  Using cached https://files.pythonhosted.org/packages/17/fd/4c2c8953a9dfe38fbe0c3adafb6355540bd98cef70cc82734acb0a4c0e2f/cryptography-2.5-cp27-cp27mu-manylinux1_x86_64.whl
Requirement already satisfied, skipping upgrade: asn1crypto>=0.21.0 in /usr/lib/python2.7/dist-packages (from cryptography) (0.24.0)
Requirement already satisfied, skipping upgrade: enum34; python_version < "3" in /usr/lib/python2.7/dist-packages (from cryptography) (1.1.6)
Requirement already satisfied, skipping upgrade: ipaddress; python_version < "3" in /usr/local/lib/python2.7/dist-packages (from cryptography) (1.0.22)
Requirement already satisfied, skipping upgrade: cffi!=1.11.3,>=1.8 in /usr/local/lib/python2.7/dist-packages (from cryptography) (1.11.5)
Requirement already satisfied, skipping upgrade: six>=1.4.1 in /usr/lib/python2.7/dist-packages (from cryptography) (1.11.0)
Requirement already satisfied, skipping upgrade: pycparser in /usr/local/lib/python2.7/dist-packages (from cffi!=1.11.3,>=1.8->cryptography) (2.19)
Installing collected packages: cryptography
  Found existing installation: cryptography 2.4.2
    Uninstalling cryptography-2.4.2:

<the agent I'm looking at has been here for 15 minutes>
```

wildlyinaccurate:
Since this breakage coincides with the release of pip 19.0, I might be more inclined to submit patch that forces the use of an older pip version (9.0?) instead.

pmeenan:
Just pushed to the release branch so all of the agents should pick it up at next reboot/start (or within the hour)

pmeenan:
If that doesn't work the best option will be to release a new image with the packages updated and just disable the auto update. It was added mostly to handle Marionette updates but that is deprecated now so I'm guessing there won't be more updates.

wildlyinaccurate:
You're so fast! I'll boot up a new agent and see how it fares with this change.

pmeenan:
Hmm, if it's failing at the cryptography install then it's not a parsing problem (other than the failed bogus lines). If it is still having issues, a blacklist or whitelist of specific modules might be needed.

wildlyinaccurate:
Yeah, this patch didn't fix it. It seems to get stuck on `pip install --upgrade` for any module that needs upgrading. I'm going to open an issue in pypa/pip.

pmeenan:
I think the problem is that the modules were originally installed as sudo and it used to silently fail before (and the code would try normal user first and then sudo). Since the modules are only upgraded on EC2 and GCE and both images were installed as sudo I changed the code to only update using sudo and I believe it is working now (testing a fresh image now to be sure).

pmeenan:
Verified that it is working now.

PR #236, opened 2019-01-23
Disable pip upgrades temporarily

wildlyinaccurate:
So #235 didn't seem to fix anything. In the mean time, I wonder if it's better to use flat-out disable the pip upgrades? At least until there's a resolution to pypa/pip#6169.

### DIFF
--- a/wptagent.py
+++ b/wptagent.py
@@ -851,7 +851,10 @@ def main():
         logging.getLogger().addHandler(err_log)
 
     if options.ec2 or options.gce:
-        upgrade_pip_modules()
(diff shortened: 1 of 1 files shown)


pmeenan:
I think the problem is that the modules were originally installed as sudo and it used to silently fail before (and the code would try normal user first and then sudo). Since the modules are only upgraded on EC2 and GCE and both images were installed as sudo I changed the code to only update using sudo and I believe it is working now (testing a fresh image now to be sure).

wildlyinaccurate:
Oh, nice! Thanks Pat :tada:

PR #239, opened 2019-02-06
Update key for Edge extensions to reflect Microsoft Webdriver changes

OleWellnitz:
There have been changes within the Microsoft Edge Webdriver. The Extension Paths are now set with a slightly different key (https://docs.microsoft.com/en-us/microsoft-edge/webdriver#w3c-webdriver). This pull request reflects these changes.

### DIFF
--- a/internal/microsoft_edge.py
+++ b/internal/microsoft_edge.py
@@ -126,7 +126,7 @@ class Edge(DesktopBrowser):
                     shutil.copy(src, extension_dir)
             except Exception:
                 pass
-        capabilities['extensionPaths'] = [extension_dir]
+        capabilities['ms:extensionPaths'] = [extension_dir]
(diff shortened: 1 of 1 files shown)


internal/microsoft_edge.py
```diff
@@ -126,7 +126,7 @@ def get_driver(self, task):
                     shutil.copy(src, extension_dir)
             except Exception:
                 pass
-        capabilities['extensionPaths'] = [extension_dir]
+        capabilities['ms:extensionPaths'] = [extension_dir]
```
pmeenan: Is there harm in setting both? If not, I'd feel better for backward compatibility if it was set for both keys.

OleWellnitz: Setting of both keys didn't cause any errors with the new Webdriver. So I don't see any problem.

PR #242, opened 2019-02-14
Hook up plumbing for marionette's (Firefox) newly-exposed buildID metadata

stephendonner:
Now that https://bugzilla.mozilla.org/show_bug.cgi?id=1525829 has landed, I'd like to be able to pull the Gecko/Firefox ```buildID``` from Marionette's Capabilities object (from the resulting test-runs' JSON file), and publish it alongside with the other metrics & metadata (```browser_version```, etc.)

@pmeenan would this be an acceptable, working change, for you?

If so, is ```fx_buildID``` fine, or would you prefer ```buildID```, ```firefox_buildID```, etc.?

/cc @davehunt @soulgalore

### DIFF
--- a/internal/firefox.py
+++ b/internal/firefox.py
@@ -38,6 +38,7 @@ class Firefox(DesktopBrowser):
         self.connected = False
         self.start_offset = None
         self.browser_version = None
+        self.browser_build = None
         self.main_request_headers = None
         self.log_pos = {}
         self.log_level = 5
@@ -149,7 +150,7 @@ class Firefox(DesktopBrowser):
             if 'browserVersion' in self.marionette.session_capabilities:
                 self.browser_version = self.marionette.session_capabilities['browserVersion']
             if 'moz:buildID' in self.marionette.session_capabilities:
-                self.fx_buildID = self.marionette.session_capabilities['moz:buildID']
+                self.browser_version = self.marionette.session_capabilities['moz:buildID']
             self.marionette.navigate(self.start_page)
             time.sleep(0.5)
             self.wait_for_extension()
--- a/internal/firefox.py
+++ b/internal/firefox.py
@@ -147,8 +147,8 @@ class Firefox(DesktopBrowser):
             self.extension_id = self.addons.install(extension_path, temp=True)
             logging.debug('Resizing browser to %dx%d', task['width'], task['height'])
             self.marionette.set_window_rect(x=0, y=0, height=task['height'], width=task['width'])
-            self.browser_version = self.marionette.session_capabilities.get['browserVersion']
-            self.browser_build = self.marionette.session_capabilities.get['moz:buildID']
+            self.browser_version = self.marionette.session_capabilities.get('browserVersion')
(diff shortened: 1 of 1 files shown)


pmeenan:
How about just 'browser_build'?  It looks like you are assigning it to a member of the python class (sort of, you need to initialize it to None in init as well) but you aren't putting it in the results anywhere. Presumably it needs to get written to the page_data.

stephendonner:
@pmeenan thanks; fixed the first few issues, and now looking into the ```page_data``` hook.

stephendonner:
I take it I also need to add an entry here, for ```browser_build```?  https://github.com/WPO-Foundation/webpagetest/blob/dbd1da1d2748678ab3b963a41085fc3077624afd/www/page_data.inc#L566

If so, do I then also just call a ```trim()``` on that field, with a bumped number (slot?)

internal/firefox.py
```diff
@@ -148,6 +149,8 @@ def launch(self, job, task):
             self.marionette.set_window_rect(x=0, y=0, height=task['height'], width=task['width'])
             if 'browserVersion' in self.marionette.session_capabilities:
                 self.browser_version = self.marionette.session_capabilities['browserVersion']
+            if 'moz:buildID' in self.marionette.session_capabilities:
+                self.browser_build = self.marionette.session_capabilities['moz:buildID']
```
davehunt: You can avoid the `if` block using `self.browser_build = self.marionette.session_capabilities.get('moz:buildID')` as `get` will return `None` if the key does not exist. Same applies to browserVersion, above.

pmeenan:
No need to update page_data.inc. That part of the code is for the legacy agent (wptdriver). The results from wptagent are loaded as a json object and whatever the agent wrote is loaded automatically (no server changes needed).

stephendonner:
@davehunt @pmeenan thanks; r?

internal/firefox.py
```diff
@@ -146,8 +147,8 @@ def launch(self, job, task):
             self.extension_id = self.addons.install(extension_path, temp=True)
             logging.debug('Resizing browser to %dx%d', task['width'], task['height'])
             self.marionette.set_window_rect(x=0, y=0, height=task['height'], width=task['width'])
-            if 'browserVersion' in self.marionette.session_capabilities:
-                self.browser_version = self.marionette.session_capabilities['browserVersion']
+            self.browser_version = self.marionette.session_capabilities.get['browserVersion']
+            self.browser_build = self.marionette.session_capabilities.get['moz:buildID']
```
davehunt: This will fail because `get` is not a dict but a method on the dict. You will need to use round brackets, such as:

```
self.browser_version = self.marionette.session_capabilities.get('browserVersion')
```

stephendonner: Fixed.

stephendonner:
Final r? @pmeenan; thx!

pmeenan:
Looks like it still isn't getting reported into the page_data results unless I'm missing something.

stephendonner:
Sorry, it's bad form of me for leaving this unimplemented, for so long; closing, and will (hopefully) revisit this later.

Issue #246, opened 2019-03-09
Windows installer


benjaminhowarth1:
There is currently no command-line installer for Windows.
I've tested & built one using Powershell & Chocolatey, and it's coming in a PR. It's just an installer for now, it does not include the batch script to kick off the Python listener to retrieve jobs, but it's a nice start.

PR #247, opened 2019-03-09
Updates #246 with a Windows PoSh installer

benjaminhowarth1:
Adds a single command-line installers for Python, .NET 3.5 Framework, and the various dependencies (git etc.). Python dependencies are installed with pip post-reboot, then a git clone to download the browser installer and wptagent. I'm working out how to pass parameters to the post-reboot script to be able to register an instance with an ID and key.


pmeenan:
Awesome, thanks.

PR #248, opened 2019-03-19
Fix parsing of submitForm command to find form correctly

saadasimosu:
The submitForm scripting command takes a single argument: the attribute
value and associated value used to identify the form in the DOM.
Currently, the logic for parsing this command attempts to use a
nonexistent (according to the syntax of the submitForm command) value to
identify the form instead of using the associated value of the
attribute. Fixed the parsing to use the appropriate value.


pmeenan:
Thanks. That'll work but the whole check for attr_value could also be removed. Looks like it was probably copied from one of the 3-parameter commands. This works though.

PR #249, opened 2019-03-20
Enable hero element times by default

wildlyinaccurate:
Accompanying PR for https://github.com/WPO-Foundation/webpagetest/pull/1261


pmeenan:
I'm hesitant to make it force-enabled in the agent.  My preference would be to control it with a config on the server (which I will enable by default on the public instance) and then have the agent honor the server's request. At least until we can assess the overhead. It should be minor but any overhead can be pretty detrimental to something like the HTTP Archive agents.

wildlyinaccurate:
Good call! I'll move it to a config on the server.

PR #251, opened 2019-03-26
Add ability to customize fallback target for network check when using Android devices

j0weiss:
We encountered a problem, when we couldn't ping `8.8.8.8` from the attached Raspis. The network check then also failed for the Android device.
We couldn't figure out why we were unable to ping this IP. `8.8.4.4` was reachable, so we temporarily used this IP address.

Adding the possibility to customize the target, gives the ability to also use domain names, which would've prevented our problem. Is there a specific reason why we ping `8.8.8.8` instead e.g. `google.com`?

### DIFF
--- a/internal/adb.py
+++ b/internal/adb.py
@@ -696,7 +696,7 @@ class Adb(object):
                             gateway = match.group(1)
                 if gateway is not None:
                     addresses.insert(0, gateway)
-                addresses.append('8.8.8.8')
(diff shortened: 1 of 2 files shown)


pmeenan:
8.8.8.8 is more widely distributed than the frontends for google.com and it is also anycasted so it generally felt safer (and didn't require a DNS lookup). Thanks for the patch though, there's no good reason it shouldn't be configurable.

PR #259, opened 2019-05-06
Add DevTools disable cache as a scripting command

saadasimosu:
Added an interface to the DevTools disable cache feature via a scripting
command. disablecache 1 disables the cache. disablecache 0 enables the
cache.


saadasimosu:
We had a case where we wanted to disable the cache to "warm up" a service worker and test other service worker functionality. Thought this might be useful to others, as well.

PR #261, opened 2019-05-09
[Lighthouse] Switch from --disable-device-emulation to --emulated-form-factor=none

wildlyinaccurate:
The `--disable-device-emulation` flag was removed in Lighthouse. `--emulated-form-factor=none` appears to have the same effect, although I wondered whether we wanted to do `--emulated-form-factor=desktop` instead?


pmeenan:
Thanks.  "none" seems safest since it is already running on a desktop and we take care of the resolution.

PR #267, opened 2019-06-24
Adding option to bypass cpu normalization

vibaldem:
Adding an option to ignore the cpu_scale_multiplier so that the throttle values don't get normalized


vibaldem:
This new option, "bypass_cpu_normalization", only applies if mobile emulation is enabled.

pmeenan:
I assume there will also be a matching webpagetest PR that allows for the test option to be passed through.

PR #270, opened 2019-07-01
Adding option to collect versions from available browsers using file …

vibaldem:
…versions for windows agents

Adding a new startup option to wptagent.py:
  --collectversion

In windows agents, this will enable collecting the browser versions using the info available in browser exes and submit them back to the controller.

### DIFF
--- a/internal/webpagetest.py
+++ b/internal/webpagetest.py
@@ -361,7 +361,7 @@ class WebPageTest(object):
         else:
             subprocess.call(['sudo', 'reboot'])
 
-    def get_test(self):
+    def get_test(self, browsers):
         """Get a job from the server"""
         import requests
         proxies = {"http": None, "https": None}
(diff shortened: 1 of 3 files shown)


internal/webpagetest.py
```diff
@@ -402,6 +402,10 @@ def get_test(self):
             uptime = self.get_uptime_minutes()
             if uptime is not None:
                 url += '&upminutes={0:d}'.format(uptime)
+            if 'browser_versions' in self.options and \
+                    len(self.options.browser_versions):
+                versions = ','.join(self.options.browser_versions)
+                url += '&browsers={0}'.format(versions)
```
pmeenan: It looks like the versions are colon delimited. I'm pretty sure that you need to url encode the versions string to be safe (even if it is working in testing). something like ...format(urllib.quote_plus(versions))

wptagent.py
```diff
@@ -901,6 +915,11 @@ def main():
             print "No browsers configured. Check that browsers.ini is present and correct."
             exit(1)
 
+    if options.collectversion and \
+            platform.system() == "Windows":
+        versions = get_browser_versions(browsers)
+        options.browser_versions = versions
```
pmeenan: It feels a little strange to be appending the version info to the options. I won't hold it back for just this but adding a 'version' entry to each entry in browsers, passing the browsers into the webpagetest instance and building the string in there feels cleaner if it's not too painful.

internal/webpagetest.py
vibaldem: Will do, thanks!

wptagent.py
vibaldem: That's actually how I had originally done it, but I wanted to minimize the changes if the option was not enabled.  But it makes more logical sense to use the browser object for browser attributes and just pass it in.

pmeenan:
Looks great, thanks.

PR #271, opened 2019-07-03
Update optimization_checks.py

Zlopi:
add new cdn provider CNAME .swiftycdn.net.


pmeenan:
Do you have info on the CDN and a site or two using it? I just need to make sure it uses distributed edges for performance and isn't just a static content server.

Zlopi:
traceroute.net.ru
melbicom.net
melbicom.ru

Anyway response header
X-CDN: SwiftyCDN

PR #274, opened 2019-07-17
No convert.exe etc in Windows 10 / newer ImageMagick installations in Windows

anordby:
ImageMagick in Windows 10 does not have convert, compare, mogrify but instead uses magick.exe. It has a convert.exe, but it does not belong to ImageMagick. https://github.com/aheckmann/gm/issues/684

So lets try to use magick.exe and not whine about it.

### DIFF
--- a/wptagent.py
+++ b/wptagent.py
@@ -65,6 +65,22 @@ class WPTAgent(object):
                                 self.image_magick['compare'] = compare
                                 self.image_magick['mogrify'] = mogrify
                                 break
+                            convert = os.path.join(path, subdir, 'magick.exe')
+                            compare = os.path.join(path, subdir, 'magick.exe')
+                            mogrify = os.path.join(path, subdir, 'magick.exe')
+                            if os.path.isfile(convert) and \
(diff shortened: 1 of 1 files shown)


vibaldem:
I thought that it does have it, but it's seeing the win10 utility c:\windows\system32\convert.exe which is for converting file systems.

anordby:
That exe file does NOT work for WPT, if magick.exe exists it should be used instead. Which my patch solves I think, it works for me.

pmeenan:
When you install ImageMagick on Windows there is an option to install the legacy exe's (convert, etc).  Does it work if you enable that during installation?  Long-term I agree that moving to magick (both on Windows and Linux) makes sense, I just need to do a LOT of testing to make sure none of the commands accidentally break, particularly for cases where multiple operations are stacked on top of each other.

thedoc31:
There's an easier way to fix this. Just bump the ImageMagick installation folder above c:\windows\system32 in the PATH statement and it'll find the ImageMagick convert.exe first instead of the Windows one.

PR #283, opened 2019-09-03
Add new CDNs

lucasRolff:
- Add QUIC.cloud
- Add header detection for hosting4cdn


lucasRolff:
@pmeenan is it possible this can get merged? hosting4cdn is present already via URL detection but now also via header detection.

quic.cloud is heavily expanding with currently 15 locations ( https://www.quic.cloud/ ) , and will add plenty more in the near future - so would be great to have it detected by webpagetest!

PR #286, opened 2019-09-17
Write long tasks found from toplevel category events to results file.

arthesh:
Long Tasks on the browser main thread can be found by processing trace events with name "ThreadControllerImpl::RunTask" having duration > 50ms. This change writes the results of these long tasks to a results file '<pathtemplate>_long_tasks.gz'. This will be used to calculate TotalLongTaskTime or other Long Tasks based metrics in post processing.

Since these task related events belong to 'toplevel' category , we include it by default when timeline or trace option is selected. We do not however save these toplevel events to the trace/timeline files unless user has explicitly added 'toplevel' category to reduce the trace sizes.

### DIFF
--- a/internal/devtools.py
+++ b/internal/devtools.py
@@ -190,7 +190,7 @@ class DevTools(object):
                 if 'type' in target and 'targetId' in target:
                     if target['type'] == 'service_worker':
                         self.send_command('Target.attachToTarget', {'targetId': target['targetId']},
-                                          wait=True)
+                                           wait=True)
 
     def close(self, close_tab=True):
         """Close the dev tools connection"""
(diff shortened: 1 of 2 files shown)


arthesh:
Closing this for now - seem to be getting intermittent exceptions on our private instance with trace parsing - will resubmit after any needed fixes.

arthesh:
Reopening this PR

pmeenan:
@arthesh did you get to the root cause of the intermittent execptions?

Did you get an estimate for how much processing overhead there is to including the toplevel category by default in the capture (even if it is thrown away)?  i.e. does it slow down the page loading and how much longer does it take to process the trace results for a site like cnn.com?

If the processing time is extended I'd feel better if it was enabled by a server config so private instances could decide to turn it on or off as needed (i.e. I wouldn't want to accidentally increase SpeedCurve's costs significantly as an unintended side-effect).

arthesh:
The exceptions were due to the some trace events not having 'dur' field - I added a check for this before reopening the PR, so this should not be an issue anymore. 

It makes sense to have a server variable that enables/disables this logic in any case - I'll push an update with the change.

PR #287, opened 2019-09-18
Extend last interactive window to the end of the run

vibaldem:
Currently the last interactive window ends at timestamp of the last timeline event.  For tests and pages which finish relatively quickly, this results in First CPU Idle (f.k.a. firstInteractive) being null.  

For this test:
https://www.webpagetest.org/result/190913_ZG_1a1fb09c7c72a2d51947801c0a99b9b2/

The test ran up to ~6.6s, but the last interactive window is only from 1405 - 3620, which disqualifies it from getting First CPU Idle.  

If there were no timeline events after the last interactive window end, I think we can assume that the page was interactive until the end of the test, so we should extend this interactive window to that time.

### DIFF
--- a/internal/support/trace_parser.py
+++ b/internal/support/trace_parser.py
@@ -476,8 +476,14 @@ class Trace():
                     len(self.interactive):
                 run_time_ms = int(math.floor(run_time * 1000.0))
                 last_window = self.interactive.pop()
-                last_window[1] = run_time_ms
-                self.interactive.append(last_window)
+                if last_window[1] != self.cpu['total_usecs']:
+                    new_window = [int(math.ceil(self.cpu['total_usecs'] / 1000.0)),
+                                  int(math.floor(run_time_ms /1000.0))]
+                    self.interactive.append(last_window)    
+                    self.interactive.append(new_window)
+                else:
+                    last_window[1] = run_time_ms
+                    self.interactive.append(last_window)
 
             # Go through all of the fractional times and convert the float
             # fractional times to integer usecs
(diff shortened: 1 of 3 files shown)


internal/support/trace_parser.py
```diff
@@ -469,6 +469,22 @@ def ProcessTimelineEvents(self):
                 self.interactive.append([int(math.ceil(self.interactive_start / 1000.0)),
                                          int(math.floor(self.interactive_end / 1000.0))])
 
+            # Extend last interactive window to the end of the run
+            # Assuming that if there are no events past the last interactive window
+            # then the page was interactive for that time
+            if self.cpu['total_usecs'] == self.interactive_end and run_time > 0 and \
```
pmeenan: Just for my sanity, can we only extend the run time if it is larger than the captured end time?  I assume it will always be the case but in case weird clock things happen I don't want to be adding a bad window to the end.

internal/support/trace_parser.py
```diff
@@ -469,6 +469,22 @@ def ProcessTimelineEvents(self):
...
+                    len(self.interactive):
+                run_time_ms = int(math.floor(run_time * 1000.0))
+                last_window = self.interactive.pop()
+                if last_window[1] != self.cpu['total_usecs']:
+                    new_window = [int(math.ceil(self.cpu['total_usecs'] / 1000.0)),
+                                  int(math.floor(run_time_ms /1000.0))]
```
pmeenan: Is this right?  usecs/1000 should result in milliseconds, not seconds.  I don't think we want to divide the run_time_ms by 1000.

PR #291, opened 2019-10-16
add lighthouseScreenshots option

connorjclark:
1. introduce `lighthouseScreenshots` option to keep screenshot thumbnails in the JSON
1. use `--skip-audits` to trim screenshot data

I didn't run any of these locally (seems daunting ...). If these changes seem suspect, LMK and I'll spent some time on that.


pmeenan:
Thanks. That should make processing a fair bit faster and less fragile.  If you actually want to be able to keep the screenshots then you'll need php-side changes to pass the test option through.

PR #294, opened 2019-10-23
fix saving lighthouse json file

connorjclark:
I messed up the logic here: https://github.com/WPO-Foundation/wptagent/commit/3c507da1e1216be6c41c5ee916d0cdbe37acdc89#diff-e2637cf67a17146071cc640a5af4bbecR547

No json is being saved anymore. this should fix it.

### DIFF
--- a/internal/devtools_browser.py
+++ b/internal/devtools_browser.py
@@ -545,10 +545,10 @@ class DevtoolsBrowser(object):
                 lh_report = None
                 with open(json_file, 'rb') as f_in:
                     lh_report = json.load(f_in)
-                if lh_report is None or 'audits' not in lh_report:
-                    with open(json_file, 'rb') as f_in:
-                        with gzip.open(json_gzip, 'wb', 7) as f_out:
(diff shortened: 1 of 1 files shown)


internal/devtools_browser.py
```diff
@@ -545,10 +545,10 @@ def run_lighthouse_test(self, task):
                 lh_report = None
                 with open(json_file, 'rb') as f_in:
                     lh_report = json.load(f_in)
-                if lh_report is None or 'audits' not in lh_report:
-                    with open(json_file, 'rb') as f_in:
-                        with gzip.open(json_gzip, 'wb', 7) as f_out:
-                            shutil.copyfileobj(f_in, f_out)
+
+                with open(json_file, 'rb') as f_in:
+                    with gzip.open(json_gzip, 'wb', 7) as f_out:
```
connorjclark: IDK if you can reuse file handles in python. I suspect no.

pmeenan:
Thanks. I had fixed the html report but didn't realize the json was also broken.

PR #298, opened 2019-11-11
Enable CPU throttling for non-mobile tests

wildlyinaccurate:
I just noticed that CPU throttling only applies when the `mobile` parameter is enabled. Is this intentional? If not, this PR allows CPU throttling to apply on non-mobile tests.

### DIFF
--- a/internal/devtools_browser.py
+++ b/internal/devtools_browser.py
@@ -112,12 +112,12 @@ class DevtoolsBrowser(object):
                 self.devtools.send_command("Emulation.setScrollbarsHidden",
                                            {"hidden": True},
                                            wait=True)
-                if (task['running_lighthouse'] or not self.options.throttle) and 'throttle_cpu' in self.job:
-                    logging.debug('CPU Throttle target: %0.3fx', self.job['throttle_cpu'])
-                    if self.job['throttle_cpu'] > 1:
-                        self.devtools.send_command("Emulation.setCPUThrottlingRate",
-                                                   {"rate": self.job['throttle_cpu']},
-                                                   wait=True)
+            if (task['running_lighthouse'] or not self.options.throttle) and 'throttle_cpu' in self.job:
+                logging.debug('CPU Throttle target: %0.3fx', self.job['throttle_cpu'])
+                if self.job['throttle_cpu'] > 1:
+                    self.devtools.send_command("Emulation.setCPUThrottlingRate",
+                                                {"rate": self.job['throttle_cpu']},
+                                                wait=True)
 
             # Location
             if 'lat' in self.job and 'lng' in self.job:


pmeenan:
Yes, it intentional because CPU is only throttled for mobile (to match the
mobile device CPU). When testing desktop or on an actual mobile device
there is no emulation needed because it is running in the desired target
environment already.

wildlyinaccurate:
Ah, right. I understand not throttling on a real device. What about wanting to test on an emulated budget Chromebook?

pmeenan:
WPT doesn't do Chromebook emulation in general.  FWIW, most of the
intel-based Chromebooks are Intel Atom/i3/i5-based and are plenty fast.
There are very few ARM ones and I'm not sure it's worth trying to emulate
their CPU perf.

wildlyinaccurate:
A budget Chromebook was just an example. What I'm interested in is measuring how being CPU-bound affects page performance, regardless of what kind of device it is. For example when I run wptagent on my laptop (an old i7 Dell XPS 13), the results are noticeably slower than the EC2 agents on webpagetest.org. Having CPU throttling for non-mobile tests would allow me to emulate loading pages in old laptops (or new laptops full of bloatware).

I did just realise something though: CPU throttling works for desktop tests on agents using cgroups. But I guess from what you're saying, this is a bug and not a feature?

pmeenan:
I don't think it does but can you double-check just to make sure we won't accidentally double-throttle with cgroups and devtools on Linux?  Also, since it is moved out of the mobile emulation block can you add a quick comment about enabling CPU throttling?

paulirish:
@wildlyinaccurate did you get a chance to make that doublecheck?

wildlyinaccurate:
Sorry, this totally fell off my radar. To be honest the use case that I had for this is no longer valid, so I'd be happy to close this.

wildlyinaccurate:
Hiiiii! I'm re-opening this because with the release of Lighthouse v6, we at SpeedCurve are hoping to tweak our CPU throttling so that our LH scores are (roughly) in line with PageSpeed. I'll also be re-testing #299 and offering that up for consideration.

@pmeenan I can confirm that Lighthouse runs are double-throttled when cgroups throttling is enabled. This is **existing behaviour** for mobile tests, which I'm guessing is unintentional. This PR would make LH runs double-throttled for all tests. Do you want me to fix the double throttling in this PR? The LH condition seems superfluous - the agent is either already throttled via cgroups, or needs to be throttled via DevTools regardless of whether it's a LH run or not. Unless I'm missing something, is the patch just:

```diff
- if (task['running_lighthouse'] or not self.options.throttle) and 'throttle_cpu' in self.job:
+ if not self.options.throttle and 'throttle_cpu' in self.job:
```

pmeenan:
@wildlyinaccurate are you SURE it is double-throttling when running lighthouse tests? As best as I can tell, start_cpu_throttling (in desktop_browser.py) is only called in on_start_recording and it is stopped in on_stop_recording. Those are only used for WPT tests during the duration of the actual video capture/test (so post-processing and trace extraction isn't throttled). None of that should be triggered for a lighthouse test.

Even with cgroups throttling for normal testing and enabled on the command-line, the devtools throttling would still be needed for lighthouse since the cgroups code path isn't activated. The PR as-is should work for enabling throttling for desktop tests (and do the correct thing for lighthouse) but as the code stands right now, it should only be throttling once for mobile tests.

wildlyinaccurate:
I can only go by what I'm seeing in the logs, which is "CPU Throttle target: n" showing up 3 times. Once to apply cgroups throttling for the normal run, then again to apply cgroups throttling for the lighthouse run, and finally to apply DevTools throttling on the lighthouse run. I tweaked the log output slightly to confirm they were coming from different code paths.

I can't seem to get ?debug=1 working on webpagetest.org to send you a test URL showing the logs.

pmeenan:
@wildlyinaccurate sorry, the messages are probably confusing.

On the cgroups side, enable_cpu_throttling() logs a debug message with the CPU target and creates the cgroup for throttling but doesn't actually start the throttling. That is run every time when Chrome is launched.

start_cpu_throttling() actually starts the throttling but doesn't output a debug message. It is only called when a WPT test is about to start recording.

disable_cpu_throttling() stops the throttling at the end of a WPT recording (before processing the result.

In the case of a lighthouse test, enable_cpu_throttling() still gets called and the empty cgroup is created for Chrome but it is left unthrottled (despite the log message) and it defers to the devtools throttling.

I just pushed an update to the log messages to hopefully make them clearer.

FWIW, the cgroups were necessary when Chrome's built-in throttling didn't work reliably. It has gotten a LOT better since then and it might make sense to just strip out all of the cgroup throttling and only use dev tools throttling (though, in theory the cgroups could be used for other browsers, practically it's only ever used for Chrome).

wildlyinaccurate:
Argh. I see now. I added extra logging in enable_cpu_throttling and totally missed start_cpu_throttling.

Is this PR still valid though? You can currently throttle CPU for desktop tests when the agent runs with `--throttle` but not without.

pmeenan:
Yeah, probably. Just want to be careful and not do it on Android devices so
it needs a check first

PR #299, opened 2019-11-11
Separate CPU throttling for lighthouse

wildlyinaccurate:
This patch adds a `throttle_cpu_lighthouse` parameter which would allow testing with separate throttling values for the "main" test and the Lighthouse test.

The context around this is that at SpeedCurve we run LH as part of the main test run, which might have no CPU throttling (e.g. an emulated iPhone X). But in order to maintain consistent LH scores, we'd like to set the CPU throttling for LH to be the same value regardless of the main test throttling.

This was branched from #298. I can submit a separate patch if that PR is invalid.

### DIFF
--- a/internal/devtools_browser.py
+++ b/internal/devtools_browser.py
@@ -112,12 +112,12 @@ class DevtoolsBrowser(object):
                 self.devtools.send_command("Emulation.setScrollbarsHidden",
                                            {"hidden": True},
                                            wait=True)
-                if (task['running_lighthouse'] or not self.options.throttle) and 'throttle_cpu' in self.job:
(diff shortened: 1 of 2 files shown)


pmeenan:
Sorry for the delay, once #298 is merged I can merge this.

wildlyinaccurate:
Hey @pmeenan, I've been thinking about what's the best way to handle this one. Do you think this `throttle_cpu_lighthouse` idea is good? Or do you think it's better to add a setting that would allow LH to do its default throttling? I.e. wptagent would not pass the `--disable-cpu-throttling` flag to LH?

pmeenan:
Maybe both as 2 different settings?

wildlyinaccurate:
No longer needed. #352 gives full control over throttling options.

PR #307, opened 2020-01-11
I've added a new cdn service provider.

muratbuyuk:
'IzHost CDN': ['.hostedcdn.net',


shaneniebergall:
@muratbuyuk appears to be a parse error here, please check

PR #309, opened 2020-01-20
Handling non-ascii characters in headers

j0weiss:
Hi Pat,

we've encountered a cookie during testing which contained non-ascii characters (german umlaut).
Unfortunately the world outside doesn't care about conventions: [Accepted cookie value characters as described at MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#Directives) 😉
With this pull request we handle this by firstly catching the thrown exception and secondly we propose to use the python `str()` object from [Python Future](http://python-future.org/str_object.html) instead of overwriting the `unicode` object by ourselves.

We couldn't create a reproducable example (our umlaut is from a customer) that we could show with the official WPT instance. Are you running your agents with Python 3 now?
We've build a simple flask example app which delivers a header with an umlaut: https://wptagent-encoding.herokuapp.com/
With our agents (running with Python 2) we got different behaviors:

- **Chrome** looses the umlaut entirely (see `Set-Cookie: Mnchen` which should read `München`):
image

- **Firefox** seems to have a similar problem:
image

If the proposed way of using `str()` from Python Future seems good to you, we're happy to fix the Firefox problem in the same way. We wanted to have a chat with you about this issue before we dig deeper into the code to find all the places where `unicode` is used.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -68,7 +68,8 @@ RUN apt-get update && \
     xvfbwrapper \
     brotli \
     'fonttools>=3.44.0,<4.0.0' \
-    marionette_driver
+    marionette_driver \
+    future
 
 COPY wptagent.py /wptagent/wptagent.py
 COPY internal /wptagent/internal
--- a/centos_install.sh
+++ b/centos_install.sh
@@ -23,7 +23,7 @@ do
 done
 sudo dbus-uuidgen --ensure
 sudo pip install --upgrade pip
-until sudo pip install dnspython monotonic pillow psutil requests ujson tornado xvfbwrapper marionette_driver
+until sudo pip install dnspython monotonic pillow psutil requests ujson tornado xvfbwrapper marionette_driver future
 do
     sleep 1
 done
--- a/docs/install.md
+++ b/docs/install.md
@@ -24,7 +24,7 @@ wptagent currently supports Windows, Linux and OSX for desktop browsers as well 
 * cgroup-tools (Linux only if mobile CPU emulation is desired)
 * traceroute (Mac and Linux)
 * Debian:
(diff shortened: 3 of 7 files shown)


pmeenan:
I'm good with the change but since future isn't a built-in module it should also be added to the various installation scripts (ubuntu_install.sh, the docker script and the windows install script) and also to [wptagent.py as a required module](https://github.com/WPO-Foundation/wptagent/blob/master/wptagent.py#L309) (so the agent will auto-install the module if it is missing).  I don't want to break existing installations.

FWIW, I'm still using 2.7 on the public instances because Firefox isn't working with marionette right now on 3 and I haven't had a chance to test Android, IE or Edge yet to make sure they are working with 3.

j0weiss:
Sounds good and thanks for the hints where we have to handle the package import.

I ran the test with the example app with Firefox (Dulles) on webpagestest.org and got a valid test result with the Cookie Header missing (whereas the test on our instance threw an exception):
image
Somehow there must be a difference between our setups. Maybe we find more informations when working further on the pull request.

j0weiss:
I've added the `future` package to the `wptagent.py` required modules and to the install scripts.
And I've tested it with Chrome on Python 2 and 3. To get the agent up and running a had to add a decode statement (see 876f92b).
With Firefox it worked with Python 2 as well and I encountered a `marionette_driver` exception when trying it with Python 3 as well. Couldn't find a bug report for the problem either.
I've also tested it on one of our agents which didn't have the package `future` preinstalled and it seemed to work fine as well.
So I think this pull request shouldn't break any existing installations.

j0weiss:
Hi @pmeenan,
I just want to ask if you think this pull request can be merged or if you see anything which should be adjusted/fixed?

pmeenan:
Looks good, thanks.  We'll land it and see how it goes ;-)

Firefox in Python 3 has issues. I haven't been opening bugs for the Python 3 port but I guess I should.  Current plan is to move Firefox off of marionette and onto webdriver - just need to find a free day sometime to do it (hopefully this weekend)

PR #325, opened 2020-02-13
Add further ffmpeg process intelligence

thedoc31:
Handles an issue where ffmpeg/gdigrab on Windows will crash if an RDP session is timed out, locked, or disconnected.
Adds a check to make sure ffmpeg is still running after it initializes, because of the previously-mentioned gdigrab issue and throw a useful error if this condition occurs.
Fixes an exception where ffmpeg crashed after initialization (again due to gdigrab issue) but desktop_browser would still try to gracefully shut it down, causing an I/O error because the pipe was already closed.
Adds a graceful ffmpeg shutdown when the video grows too large.
Removes unused 'errno' import.

### DIFF
--- a/internal/desktop_browser.py
+++ b/internal/desktop_browser.py
@@ -497,14 +497,15 @@ class DesktopBrowser(BaseBrowser):
                     # gdigrab will fail to capture if Secure Desktop and UAC are enabled
                     # or an RDP session is minimized/disconnected/locked
                     if self.ffmpeg.poll() is None:
-                        logging.debug("Video capture process is still running...")
+                        logging.debug('Video capture process is still running...')
                         self.video_capture_running = True
                     else:
                         self.video_capture_running = False
-                        logging.exception("ERROR: ffmpeg started but gdigrab did not provide a stream.")
-                        raise RuntimeError('Video could not be captured with this run. Is UAC enabled or RDP minimized/disconnected/locked?')
+                        self.task['error'] = 'ffmpeg started but gdigrab did not provide a stream.'
+                        logging.exception('ERROR: %s', self.task['error'])
                 except Exception:
-                    logging.exception('Error starting video capture')
+                    self.task['error'] = 'Video capture failed to start.'
+                    logging.exception('ERROR: %s', self.task['error'])
             # start the background thread for monitoring CPU and bandwidth
             self.usage_queue = multiprocessing.JoinableQueue()
             self.thread = threading.Thread(target=self.background_thread)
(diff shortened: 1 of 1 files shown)


thedoc31:
@pmeenan only question I have on this is if a video capture fails and the test is set to capture video, should the test even proceed? With this fix, the test will proceed but video will appear as 'not available', and all the metrics that go with it won't show. It seems to me like a better fix for this might just be to fail the entire test with an error like "Video capture couldn't be started" instead. Thoughts?

internal/desktop_browser.py
```diff
@@ -514,19 +523,22 @@ def on_stop_capture(self, task):
...
-        if self.ffmpeg is not None:
+        if self.ffmpeg is not None and self.video_capture_running:
             logging.debug('Stopping video capture')
             self.video_capture_running = False
             if platform.system() == 'Windows':
                 logging.debug('Attempting graceful ffmpeg shutdown\n')
                 self.ffmpeg.communicate(input='q')
                 if self.ffmpeg.returncode is not 0:
                     logging.exception('ERROR: ffmpeg returned non-zero exit code %s\n', str(self.ffmpeg.returncode))
+                    raise RuntimeError('ffmpeg encountered an error during exit. Video may be corrupt.')
```
pmeenan: I'd rather not re-raise the exception without knowing how far up it will go before getting caught.  It may leave other things like traffic-shaping in a bad state.

internal/desktop_browser.py
```diff
@@ -485,17 +484,27 @@ def on_start_recording(self, task):
...
+                    # "Output file #0 does not contain any stream"
+                    # gdigrab will fail to capture if Secure Desktop and UAC are enabled
+                    # or an RDP session is minimized/disconnected/locked
+                    if self.ffmpeg.poll() is None:
+                        logging.debug("Video capture process is still running...")
+                        self.video_capture_running = True
+                    else:
+                        self.video_capture_running = False
+                        logging.exception("ERROR: ffmpeg started but gdigrab did not provide a stream.")
+                        raise RuntimeError('Video could not be captured with this run. Is UAC enabled or RDP minimized/disconnected/locked?')
```
pmeenan: Rather than re-raising an exception, if we want to mark the test as failed we should be able to set the error in self.task['error'].

internal/desktop_browser.py
```diff
@@ -564,10 +576,14 @@ def on_stop_recording(self, task):
...
-            logging.debug('Waiting for video capture to finish')
+        if self.ffmpeg is not None and self.video_capture_running:
+            logging.debug('Waiting for video capture to finish...')
             self.ffmpeg.communicate(input='q')
-            self.ffmpeg = None
+            if self.ffmpeg.returncode is not 0:
+                logging.exception('ERROR: ffmpeg returned non-zero exit code %s\n', str(self.ffmpeg.returncode))
+            else:
+                logging.debug('ffmpeg shutdown gracefully\n')
+                self.ffmpeg = None
```
pmeenan: Even in the non-zero return case we should be setting self.ffmpeg to none.

internal/desktop_browser.py
```diff
@@ -737,7 +753,12 @@ def background_thread(self):
                     logging.debug('Stopping video capture - File is too big: %d', video_size)
                     self.video_capture_running = False
                     if platform.system() == 'Windows':
-                        os.kill(self.ffmpeg.pid, signal.CTRL_BREAK_EVENT) #pylint: disable=no-member
+                        self.ffmpeg.communicate(input='q')
+                        if self.ffmpeg.returncode is not 0:
+                            logging.exception('ERROR: ffmpeg returned non-zero exit code %s\n', str(self.ffmpeg.returncode))
+                        else:
+                            logging.debug('ffmpeg shutdown gracefully\n')
+                            self.ffmpeg = None
```
pmeenan: same

thedoc31:
Ok. I tweaked the code so we won't re-raise an error, and we'll fail the test if video capture starts but dies. In any case when ffmpeg is stopped the self.ffmpeg will now be set to None. Thanks for the input!

pmeenan:
Awesome, thanks.  Looks great and thanks for refactoring to make the shutdown cleaner.

sejoker:
Hello @thedoc31 @pmeenan. I've `git bisect` the issue of `not available` video on *Ubuntu* running all versions of Chrome, e.g. https://www.webpagetest.org/result/200224_JH_ea5ddb1b5c3acbdd66986208969c4070/. It's a serious regression IMO and requires quite urgent attention.

pmeenan:
@sejoker I reverted and pushed a new release. We can re-apply the windows-specific fixes with a windows-only check once we're sure it is in a good place again.

sejoker:
Thanks, Pat. It works fine now on the latest master.

PR #329, opened 2020-02-20
Fix close_browser function, set wait() for gracefull closing

trimon90:
When the agent is launched in docker and close_browser () is called with kill_all (), this creates a memory leak. Added wait () for the normal completion of the process.

### DIFF
--- a/internal/desktop_browser.py
+++ b/internal/desktop_browser.py
@@ -292,18 +292,23 @@ class DesktopBrowser(BaseBrowser):
         """Terminate the browser but don't do all of the cleanup that stop does"""
         if self.proc:
             logging.debug("Closing browser")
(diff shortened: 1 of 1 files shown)


internal/desktop_browser.py
```diff
@@ -293,9 +293,9 @@ def close_browser(self, job, _task):
         if self.proc:
             logging.debug("Closing browser")
             try:
-                self.proc.wait(timeout=15)
-            except TimeoutExpired:
-                self.proc.wait(timeout=15)
```
Mywebstoreshopping: self.proc.wait(timeout=15)
            except TimeoutExpired:
                self.proc.wait(timeout=15)

PR #330, opened 2020-02-24
Revert "Add further ffmpeg process intelligence"

pmeenan:
Reverting first and we can re-apply bits when we are sure it doesn't affect Ubuntu video capture.

Reverts WPO-Foundation/wptagent#325


thedoc31:
Will check to see if I can figure out what went wrong. Our checks on Ubuntu have been unaffected after this change, as far as I can tell.

PR #352, opened 2020-05-28
Allow custom Lighthouse throttling via config path

wildlyinaccurate:
So this is sort of the "give me full control of Lighthouse" approach to #299. If you provide a config path (for example `/usr/lib/node_modules/lighthouse/lighthouse-core/config/lr-mobile-config.js`) the agent will forgo any traffic shaping or CPU throttling, leaving it entirely up to Lighthouse to do whatever is in the config.

The downside to this is the simulated throttling that Lighthouse does is not great. The upside is that you can have full control over how WPT runs LH, and wptagent doesn't need a bunch of separate LH options for controlling network & CPU throttling independently.

The motivation behind this is that SpeedCurve gets a lot of questions about why the LH scores in SC are different to the LH scores in PageSpeed Insights. To avoid that, we'd rather just use the exact same config that PSI uses.


internal/devtools_browser.py
```diff
@@ -584,15 +589,16 @@ def run_lighthouse_test(self, task):
...
-                       '--disable-cpu-throttling',
-                       '--throttling-method', 'provided',
                        '--enable-error-reporting',
                        '--max-wait-for-load', str(int(time_limit * 1000)),
                        '--port', str(task['port']),
                        '--output', 'html',
                        '--output', 'json',
                        '--output-path', '"{0}"'.format(output_path)]
+            if self.job['lighthouse_config_path']:
+                command.extend(['--config-path', self.job['lighthouse_config_path']])
```
pmeenan, line 599: Do you need to do any sanitizing here or are you going to make absolutely sure that it is only ever statically defined in the server? Otherwise you can get arbitrary code execution on the agents.

wildlyinaccurate: Ah, good point. Do you think a simple `os.access(path, os.R_OK)` would suffice?

wildlyinaccurate:
I haven't verified that it matches the PSI scores. I was going to hang out until they upgrade to LH v6 but I guess I could downgrade to LH v5 on my test agent. I was going by [this tweet](https://twitter.com/cjamcl/status/1265785207027712003) but it's kind of irrelevant exactly which configs they use. The point is that we'd be able to run any arbitrary config rather than being limited to a subset of LH options that WPT exposes as job settings.

I'm aware that this PR is pretty overkill though. Really what we're after is being able to independently control the network and CPU throttling for Lighthouse, regardless of the throttling used for the "parent" test. I guess we could run a totally separate Lighthouse-only test, but it might be difficult to consolidate that with the main test run.

internal/devtools_browser.py
pmeenan: It looks like you got rid of the disable cpu and network throttling in the "else" case. Was that intentional?

As far as sanitizing, a couple of options come to mind:

1 - Store the configs on the server and pass the json down as part of the test. The agent could write the config to the temp path and there's no risk of execution. That also keeps you from having to build new agent images every time the config changes.

2 - Keep the configs in a well-known location and only let the test specify the name of the config file to use. Then you can make sure it's only alphanumeric characters.

wildlyinaccurate: Do you mean the `--disable-{network,cpu}-throttling` flags? Those were removed a couple of versions ago in favour of `--throttling-method`.

I'll go with your first suggestion and have the full config JSON passed to the agent. This requires slightly more work for the test runner (can't just point at known LH configs) but agree it's way safer.

wildlyinaccurate:
Latest commit takes a `lighthouseConfig` option as a JSON string (validated by `json.loads()`)

PR #359, opened 2020-06-26
New param `shaperLimit` to runtest.php for network shaping

gfx:
`tc qdisc` can take `limit` parameter for the buffer size of the queue but there's no parameter to set it. If I understand correctly, no "correct" value for `limit` can be calculated so it should be configured by test designers. 

The `limit` param is described in [`netem(8)` manpage](https://www.man7.org/linux/man-pages/man8/tc-netem.8.html):

>  limit packets
>       maximum number of packets the qdisc may hold queued at a time.

---

I'll test this patch on our private instance, and at the same time, I have made this pull-req here to discuss this feature.

## refs

* https://github.com/WPO-Foundation/webpagetest/pull/1367

### DIFF
--- a/internal/traffic_shaping.py
+++ b/internal/traffic_shaping.py
@@ -91,9 +91,9 @@ class TrafficShaper(object):
         plr = .0
         if 'plr' in job:
             plr = float(job['plr'])
-        tcQdiscLimit = 0
-        if 'tcQdiscLimit' in job:
-            tcQdiscLimit = self._to_int(job['tcQdiscLimit'])
+        shaperLimit = 0
+        if 'shaperLimit' in job:
+            shaperLimit = self._to_int(job['shaperLimit'])
         if self.shaper is not None:
             # If a lighthouse test is running, force the Lighthouse 3G profile:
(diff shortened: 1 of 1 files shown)


pmeenan:
Patch looks good to me if it works correctly in your local testing.

Only caveat I saw was that the default limit is in packets but can be bytes with bfifo. Packets is probably good enough but I am not sure what networking gear tends to use for their buffers. That's probably worth a separate option in the future if it matters.

Might be worth re-naming it to shaperLimit and also using it on Windows to set the in/out buffers for winshaper: https://github.com/WPO-Foundation/win-shaper/blob/master/exe/exe.cpp#L267

gfx:
Thanks!

The set of patches (wptagent & webpagetest server) works on my local env, and also renamed `s/tcQdiscLimit/shaperLimit/`. 

The `shaperLimit` param is not used in other shapers yet, but I think this PR is good enough for time being.

And as you said, the unit of `shaperLimit` is packets and it might be inconvenient , but if it accepts bytes in the future, we can use "b" suffix to distinguish its unit, jus like as `tc qdisc`  shows. So, I keep this PR as is so far.

PR #364, opened 2020-07-13
Skip invalid fonts to prevent exceptions

wildlyinaccurate:
An alternative fix for #363.


pmeenan:
I like this approach better.

PR #366, opened 2020-07-20
Disable cgroups throttling on Lighthouse runs when lighthouseConfig is provided

wildlyinaccurate:
When I submitted #352 I accidentally enabled double CPU throttling for Lighthouse runs when a) cgroups throttling is enabled and b) the lighthouseConfig option is used. My test agents all run in Docker containers so I never actually tested how the lighthouseConfig worked on an agent that runs with the `--throttle` flag. The cgroups throttling is still applied and then Lighthouse applies its own DevTools throttling on top of that.

I don't know if this patch is the "right" approach - I wasn't sure whether `WPTAgent` should be concerned with job parameters like `lighthouseConfig`. I guess the alternative would be to put this logic in `DesktopBrowser.enable_cpu_throttling()` and have it set `throttling_cpu = False`. LMK what you think.

Edit: Also keen to know your thoughts on comments in source code. This block of code is a bit like "huh? what?" and could probably do with a comment explaining the logic, but I didn't want to pollute your codebase with comments.

### DIFF
--- a/internal/desktop_browser.py
+++ b/internal/desktop_browser.py
@@ -282,7 +282,9 @@ class DesktopBrowser(BaseBrowser):
 
     def launch_browser(self, command_line):
         """Launch the browser and keep track of the process"""
-        command_line = self.enable_cpu_throttling(command_line)
+        if not self.job['lighthouse_config'] or not self.task['running_lighthouse']:
+            # Only enable CPU throttling for Lighthouse runs if a custom config was not provided
+            command_line = self.enable_cpu_throttling(command_line)
         logging.debug(command_line)
         if platform.system() == 'Windows':
             self.proc = subprocess.Popen(command_line, shell=True)


pmeenan:
I'm a bit concerned about setting it at the job level.  I'd be more comfortable if it is done at the task level and then the task just re-copies it from the job each time (and the throttle code checks for it at the task instead of job).

The main reason is if there are multiple runs and the test is not sharded, the lighthouse test is run after the first run but subsequent runs will use the same job object and result in throttling being disabled.

wildlyinaccurate:
Thanks for merging! I was meant to write a comment yesterday to assure you that I'd tested this in various scenarios and have logs from each run :)

wildlyinaccurate:
Huh, I'm not sure this actually fixed anything on our EC2 agents. What I was observing before this PR was that tests with relatively high CPU throttling (I was testing with throttle_cpu=3.2) had much lower CPU/memory scores in Lighthouse than unthrottled tests. This usually correlated with lower LH performance scores.

Here are some sample tests with this patch included:

- [This test](https://wpt1.speedcurve.com/lighthouse.php?test=200721_BZ_5fd8fdc7e6a49eb41998be5912f75105&run=1) was run with throttle_cpu=0.6 (which ended up being 0.586 after applying the multiplier). Lighthouse gave it a CPU/memory score of 1324.
- [This test](https://wpt1.speedcurve.com/lighthouse.php?test=200721_88_76504b46db0bf9c5345caf59cb66cbd3&run=5) used throttle_cpu=1.1 (actual: 1.078) and got a CPU/memory score of 954.
- [This test](https://wpt.speedcurve.com/lighthouse.php?test=200721_HA_726131793a4efa1e0e7ca0fb244d520f&run=5) used throttle_cpu=3.2 and got a CPU/memory score of 968

The last two tests seem to be working as expected: regardless of the throttle_cpu value, their CPU/memory scores are pretty close to each other. The first test though, with a target throttle of <1, has a CPU/memory score 300-400 points higher. Obviously the main difference between these tests is that the last two run Chrome via `cgexec`. Do you think this is introducing overhead?

pmeenan:
It's worth checking but throttle values < 1 disable throttling (no way to
make it go faster than native). Should check that cgroups aren't used at
all in that case

wildlyinaccurate:
Yeah, the logs indicate that cgroups aren't used at all when throttle is <1. The thing that surprised me is that even when the throttle value is very close to 1 (1.078), there's a significant effect on CPU performance.

I guess the real issue here is that my patch doesn't seem to be preventing `enable_cpu_throttling` from being called on our EC2 agents. BUT there's a wider issue that even with `cpu.cfs_period_us=1000` and `cpu.cfs_quota_us=1000` (which means "no throttling", right?) there seems to be a significant impact on CPU performance.

Am I just misplaced in believing that throttle_cpu=1 should have (roughly) the same performance as not specifying throttle_cpu at all?

pmeenan:
The chrome throttling got a lot better a few years ago. Might be worth just
disabling the cgroups usage entirely and just use the devtools throttling.

pmeenan:
I just yanked out all of the cgroup logic and changed it to always use devtools for all platforms.  That should be a lot cleaner and make sure that lighthouse and WPT are both using the same throttling.

Godspeed2021 (approved):
http://www.w3.org/2005/Atom

Dustin4444:
✨✨ Here's [an AI-assisted sketch of how you might approach this issue](https://copilot-workspace.githubnext.com/catchpoint/WebPageTest.agent/pull/366?shareId=698e1a78-2c6b-4579-b1c7-70df62e017d8) saved by @Dustin4444 using Copilot Workspace v0.27

PR #383, opened 2020-10-07
Enable support Safari browser via webdriver on MacOS

wuwang-wang:
Signed-off-by: Arthur Wang <wuwang@google.com>


internal/safari_webdriver.py
```diff
@@ -0,0 +1,1144 @@
...
+        self.webdriver.set_window_rect(x=0, y=0, height=height, width=width)
+
+    def disconnect_driver(self):
+        """Disconnect Webdriver"""
+        if self.extension_id is not None and self.addons is not None:
+            try:
+                self.addons.uninstall(self.extension_id)
+            except Exception:
+                logging.error('Error removing addons')
+            self.extension_id = None
```
pmeenan, line 255: It doesn't look like extension_id is used anywhere in the code.  Is this left over?

pmeenan: Same goes for self.addons

internal/safari_webdriver.py
```diff
@@ -0,0 +1,1144 @@
...
+
+    def close_browser(self, job, task):
+        """Terminate the browser but don't do all of the cleanup that stop does"""
+        self.connected = False
+        self.disconnect_driver()
+        DesktopBrowser.close_browser(self, job, task)
+        # make SURE the Firefox processes are gone
+        if platform.system() == "Linux":
+            subprocess.call(['killall', '-9', 'safari'])
+            # subprocess.call(['killall', '-9', 'firefox-trunk'])
```
pmeenan, line 269: Remove the firefox commented-out code

internal/safari_webdriver.py
```diff
@@ -0,0 +1,1144 @@
...
+            with io.open(os.path.join(self.script_dir, 'hero_elements.js'), 'r', encoding='utf-8') as script_file:
+                hero_elements_script = script_file.read()
+            script = hero_elements_script + '(' + json.dumps(custom_hero_selectors) + ')'
+            hero_elements = self.execute_js(script)
+            if hero_elements is not None:
+                path = os.path.join(task['dir'], task['prefix'] + '_hero_elements.json.gz')
+                with gzip.open(path, GZIP_TEXT, 7) as outfile:
+                    outfile.write(json.dumps(hero_elements))
+
+    def process_message(self, message):
```
pmeenan, line 479: Is all of the message handling identical to the iOS agent?  If so it would be awesome if we could refactory them both into a shared parent class that both use (or for now add a todo comment to refactor)

internal/safari_webdriver.py
```diff
@@ -0,0 +1,1144 @@
...
+                'about:blank']
+        if self.path.find(' ') > -1:
+            command_line = '"{0}"'.format(self.path)
+        else:
+            command_line = self.path
+        command_line += ' ' + ' '.join(args)
+        command_line = self.disable_fsync(command_line)
+        from selenium import webdriver
+        capabilities = webdriver.DesiredCapabilities.SAFARI.copy()
+        self.webdriver = webdriver.Safari(desired_capabilities=capabilities)
```
pmeenan, line 141: Is the extension installed at any point or is it manually installed in the system globally?

pmeenan (commented):
Thanks for doing this, it looks great.  Main question I had was around the extension.  There are a lot of xcode workspace files for building the extension and I don't see either a binary for the extension or any way that it gets installed at runtime.  Is the extension used?  If so, does it need to be installed manually on the test system?

wuwang-wang:
Sorry for the delay. I was running webpagetest for 450 runs last week and
it occupied my macOS machine. I am going to address the comments
today/tomorrow.
Thanks!

--
Arthur Wang
SWE @Clank

wptagent.py
```diff
@@ -432,6 +436,8 @@ def startup(self, detected_browsers):
                 print("Error configuring traffic shaping, make sure secure boot is disabled.")
             else:
                 print("Error configuring traffic shaping, make sure it is installed.")
+            # TODO(wuwang): Remove this comments if pfctl/dnctl works,
```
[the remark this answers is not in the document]
wuwang-wang: Done

internal/safari_webdriver.py
wuwang-wang: done

internal/safari_webdriver.py
wuwang-wang: done

internal/safari_webdriver.py
wuwang-wang: I do not think we are using same logic. I recall iOS was using web inspector protocol. I will have another look and see if there are logic that could be merged, module being reduced, maybe some refractory work could be done.

internal/safari_webdriver.py
wuwang-wang: That extension. I was dev signed and install manually. I want to get it official signed by apple, and then I could figure out how to install it via script. Right now, because it is dev signed, I have to manually install. So I did not bother to figure that out yet.

PR #387, opened 2020-12-08
Fix a video resolution issue when devicePixelRatio is not 1.0

wuwang-wang:


pmeenan:
I changed the logic to get the screen DPR directly from MacOS instead of through the browser. That should also work for the case where we are using the iOS simulator where the screen and device DPR will be different.

https://github.com/WPO-Foundation/wptagent/commit/1cae0fad00536d06cb51382d2c42691c471834f6

Let me know if that doesn't solve it.

PR #389, opened 2020-12-09
Replace a few tab with spaces inside desktop_browser.py

wuwang-wang:


pmeenan:
Try pulling the latest.  I believe I did this last night manually.

pmeenan:
Why are you removing the Firefox MOZ_LOG from Firefox with a commit comment about removing it for Android?  The MOZ_LOG env variable is needed for Firefox to get the logs which are parsed for the request details (or did you not mean to update the PR?)

wuwang-wang:
Sorry. that was a typo. I was meant for "firefox desktop", not "firefox
android".

This is due to

   - I ran into Selenium.InvalidArgumentException for complaining there is
   no field 'env' in mz.firefoxOptions
   - the change log:
   https://searchfox.org/mozilla-central/source/testing/geckodriver/CHANGES.md
   was saying env was added, but only for firefox android.

I was running latest firefox on MacOS 11.1
Thanks!

--
Arthur Wang
SWE @Clank

pmeenan:
Did you mean to create a new PR or update the old one?  If this is where you want it, can you update the title and description to explain why the db is being copied?

wuwang-wang:
Oh. We observe some firefox page load section has a download section of revoked certificate list, which starts right before the rest of page load section. It adds some addition latency on this. Some further study finds that those cert9.db are stored in the profile directory, and will be wiped between rendering sections. So I decide to backup cert9.db after a run, and copy the cert9.db right before a new run. This will help avoid seeing a "download of revoked certificate list" section. 

I am running 200 runs right now to confirm that behavior.

pmeenan:
Can you send me an example test where the list is downloaded?  We're probably better off just blocking the download so it doesn't even happen the first time.  The browser is updated every 6 weeks anyway and any sites that check revocations will usually do it directly with the CA - I assume this is the emergency manual list.

wuwang-wang:
Here it is:
http://34.94.181.102/result/210208_V2_7711a2450e8fd268908b661979cc0933/

See the http://ocsp.pki.goog/gts1o1core request in this waterfall
<https://www.google.com/url?sa=D&q=http%3A%2F%2F34.94.181.102%2Fresult%2F210208_V2_7711a2450e8fd268908b661979cc0933%2F>
for
example. This adds several RTTs in latency to the main frame request. We
suspect this happens because WPT clears Firefox's data between runs and so
it has to check for new certificate revocations on startup. That's very
unlikely to be fair to Firefox and is likely just an artifact of WPT.

Thanks!

--
Arthur Wang
SWE @Clank

wuwang-wang:
The site is an internal hosted site with 1MB data to download. Other
than that, the site is just a static content site. We are using the site to
identify what kind of noise the infra could bring up and find a way to
mitigate them.

Thanks!

--
Arthur Wang
SWE @Clank

pmeenan:
That is not a CRL download, that is an online OCSP verification check for a single certificate (similar use cases but very different).

Firefox does OCSP validation for all certificates - EV and DV.  Chrome only does it for EV certificates.

It will only be cached for Firefox if a user has been to that specific origin before.

In most cases the OCSP check can be avoided if the server [staples the revocation check](https://hpbn.co/transport-layer-security-tls/#ocsp-stapling) but that doesn't [appear to happen in this case](https://www.ssllabs.com/ssltest/analyze.html?d=probe.googlezip.net&s=142.250.72.206) (and intermediary certificates can't be stapled).

In this specific case, the revocation check is real and does slow down the connection.

wuwang-wang:
In that case, put user_pref("security.ssl.enable_ocsp_stapling", disable);
into prefs.js should work. but it does not seem so. Let me poke around it a
little bit more
Thanks!

--
Arthur Wang
SWE @Clank

pmeenan:
Disabling ocsp stapling wouldn't disable the ocsp validation check, it would disable support for stapling if the certificate had a stapled response included (which it doesn't).

wuwang-wang:
What of also adding user_pref("security.OCSP.enabled", 0);
Thanks!

--
Arthur Wang
SWE @Clank

pmeenan:
I wouldn't recommend doing that globally but you can try it with a test script using the "firefoxPref" script command.

wuwang-wang:
Would like to know more about that part:
>
>  try it with a test script using the "firefoxPref" script command.
>
Thanks!

--
Arthur Wang
SWE @Clank

pmeenan:
The [WebPageTest scripting](https://docs.webpagetest.org/scripting/#firefoxPref) lets you set arbitrary firefox prefs as part of a test.  i.e.

```
firefoxPref	security.OCSP.enabled	0
navigate	%URL%
```

If you paste that into the script tab of advanced settings (or pass as an urlencoded param for "script" then it should set the pref as part of the test before loading the page (%URL% is a substitution that will automatically be replaced by the URL being tested).

wuwang-wang:
Thanks! I will give it a try and see if we could send that firefoxPref
script over REST API.
Thanks!

--
Arthur Wang
SWE @Clank

pmeenan:
Through the API it would be script=firefoxPref%09security.OCSP.enabled%090%0Anavigate%09%25URL%25

wuwang-wang:
OK. I have it working with script. Let me revert this change tomorrow.
Thanks!

--
Arthur Wang
SWE @Clank

pmeenan:
Closing as it no longer matches the initial work and has a bunch of separate changes in it. If any of those should be landed please open separate PRs for each of the changes.

PR #405, opened 2021-02-26
Add flag to disable optimisation guide hints on a Moto G4

Nooshu:
for #404


pmeenan:
Whoops. We landed at the same time.  Thanks for the PR though!

PR #406, opened 2021-03-19
Adding JuraganCDN as CDN provider

midorinet:
Adding Indonesian CDN called JuraganCDN


pmeenan:
Do you happen to have a link to a network map for Juragan CDN or a test page that I can hit from a few different locations to make sure it is not all served from the same location?

midorinet:
Hi

You can check on https://www.kaskus.co.id

Assets for CDN is s.kaskus.id

Issue #395, opened 2021-01-19
Run a lighthouse test against the current browser session


M-Z:
I can see the comment "Run a lighthouse test against the current browser session"
But, after launching the agent with the browser GUI, I can see the browser closes and a new one spawns just for LH. Which loses the cookies/storage/etc. for my tests.
I applied a quick fix (don't know if this is the right approach here) that might help

wptagent.py
L#129
`self.task['running_lighthouse'] = False
     if self.task['run'] == 1 and not self.task['cached'] and \
         self.job['warmup'] <= 0 and \
         self.task['error'] is None and \
         'lighthouse' in self.job and self.job['lighthouse']:
                 if 'page_result' not in self.task or \
                     self.task['page_result'] is None or \
                     self.task['page_result'] == 0 or \
                     self.task['page_result'] == 99999:
                         self.task['running_lighthouse'] = True
              self.run_single_test()
              self.wpt.get_bodies(self.task)
              elapsed = monotonic() - start
              logging.debug('Test run time: %0.3f sec', elapsed)`
L#191
`if browser is not None:
            browser.prepare(self.job, self.task)
            browser.launch(self.job, self.task)
            try:
                browser.run_task(self.task)

                if self.task['running_lighthouse']:
                    self.task['lighthouse_log'] = \
                        'Lighthouse testing is not supported with this browser.'
                    try:
                        browser.run_lighthouse_test(self.task)
                    except Exception:
                        logging.exception('Error running lighthouse test')
                    if self.task['lighthouse_log']:
                        try:
                            log_file = os.path.join(self.task['dir'], 'lighthouse.log.gz')
                            with gzip.open(log_file, GZIP_TEXT, 7) as f_out:
                                f_out.write(self.task['lighthouse_log'])
                        except Exception:
                            logging.exception('Error compressing lighthouse log')
`

devtools_browser.py
L#54
Removed the `ret = self.devtools.wait_for_available(self.CONNECT_TIME_LIMIT)`
and made it just connect once to the dev tools interface and not wait/reconnect to the devtools
L#267
`if not self.task['running_lighthouse']:
                self.task = None`

pmeenan:
I don't understand exactly what you are trying to do.  You can't re-use a browser session across a WPT test and a Lighthouse test.

Lighthouse makes 3-4 passes across a page measuring different things and is expecting to be in full control and starting from a clean slate.  If you need to specify additional headers or cookies it needs to be done explicitly.

[...]

PR #420, opened 2021-07-06
lighthouse for current browser session

imkalvin:
Hi @pmeenan,

In my org, we need to test against pages behind authentication (which was identified with a cookie with max-age=session). With these changes, WPT is now able to do that (tested on Windows+Ubuntu 18.04).

Pls help to take a look whether this is something WPT wants to take in.

This might also address the issue https://github.com/WPO-Foundation/wptagent/issues/395

### DIFF
--- a/internal/devtools_browser.py
+++ b/internal/devtools_browser.py
@@ -744,6 +744,7 @@ class DevtoolsBrowser(object):
         proc.communicate()
 
     def run_lighthouse_test(self, task):
+        self.task = task
         if self.must_exit_now:
             return
         self.profile_start('dtbrowser.run_lighthouse_test')
--- a/wptagent.py
+++ b/wptagent.py
@@ -142,8 +142,14 @@ class WPTAgent(object):
                         self.task = self.wpt.get_task(self.job)
                         while self.task is not None:
                             start = monotonic()
+
+                            self.task['running_lighthouse'] = False
+                            self.browser = self.browsers.get_browser(self.job['browser'], self.job)
+                            if self.browser is not None:
+                                self.browser.prepare(self.job, self.task)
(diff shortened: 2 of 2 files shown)


pmeenan:
It might be better if the capability was an option so it could be enabled explicitly when desired.  Otherwise there is no way to get a "clean" lighthouse test with no cookies or cache populated.

imkalvin:
> It might be better if the capability was an option so it could be enabled explicitly when desired. Otherwise there is no way to get a "clean" lighthouse test with no cookies or cache populated.

Then I assume we will need to enhance the server part to add a new option, then pass along to agents...right?

pmeenan:
Either an explicit option or only use this mode when a script is provided.  I'm happy to plumb the test setting if you're worried about the php side of things (it's just index.php and runtest.php and can be copied from some of the other options)

imkalvin:
> Either an explicit option or only use this mode when a script is provided. I'm happy to plumb the test setting if you're worried about the php side of things (it's just index.php and runtest.php and can be copied from some of the other options)

Thanks @pmeenan ! 

I think using this mode when  a script is provided will make better sense. I will update the PR later.

imkalvin:
Hi @pmeenan ,

I had wrapped the new logic with a flag check upon task['script']. Pls help to review and proceed. Thanks!

M-Z:
Hi,
I gave this a try and it worked as it should for me.
Would be great if this is confirmed to patch my agents with it.

eugeneidoronin:
If it works that would be great to have it merged. Thank you.

pmeenan:
I need to double-check but I think the non-script case may also trigger it because a simple URL test is also converted into a 1-step navigate script.  I think the reuse check need to check that the script length > 1 instead of > 0

PR #431, opened 2021-09-14
Map netlog priorities to the Dev Tools equivalents

pmeenan:
This changes the reported Priorities to use the same names as Dev Tools uses for the different priority levels.  I am in the process of moving Blink to also use the dev tools names so hopefully we can have one set of priority names that we use when talking about Chrome prioritization (netlog internally will still use the net names because it is locked into the API but I will change the [priorities doc](https://docs.google.com/document/d/1bCDuq9H1ih9iNjgzyAL0gpwNFiEP4TZS-YLRp_RuMlc/edit?usp=sharing) to de-emphasize the net names).

With this, both Dev Tools and WPT will show the same names for priorities which should make it a lot easier to talk about, particularly as Priority Hints are going to be coming along soon.

I'll have a PR shortly to also map the names in the UI for older tests so they are displayed consistently.

### DIFF
--- a/internal/support/devtools_parser.py
+++ b/internal/support/devtools_parser.py
@@ -51,6 +51,13 @@ class DevToolsParser(object):
         self.result = {'pageData': {}, 'requests': []}
         self.request_ids = {}
         self.script_ids = {}
+        self.PRIORITY_MAP = {
+            "HIGHEST": "Highest",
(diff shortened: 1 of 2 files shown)


tkadlec:
Ooo...yeah, I like this a lot. That was always one of those things that seemed like an arbitrary difference.

PR #436, opened 2021-09-29
Fix #435 by adding a break 

### LINKED ISSUE #435: Firefox tests don't accurately depict duplicate requests (service worker)
When testing a site in Firefox where:

- a service worker exists
- the service worker requests a resource that either matches the main document, or has already been loaded by the main document

The waterfall incorrectly combines the two requests leading to some very weird waterfalls. Example, request 24 in the below waterfall for the main document.

bad-waterfall-ff

tkadlec:
Looks like issue #435 was caused by an over eager set of for loops that simply needed a `break` when a match occurred so that the same request couldn't be matched to multiple requests in the logs.

### DIFF
--- a/internal/firefox.py
+++ b/internal/firefox.py
@@ -1081,6 +1081,7 @@ class Firefox(DesktopBrowser):
                             and 'start' in req and request['full_url'] == req['url']:
                         req['claimed'] = True
                         self.populate_request(request, req)
+                        break
                 except Exception:
                     logging.exception('Error populating request')
         # Add any events from the logs that weren't reported by the extension


scottjehl:
aw. we all need a break sometimes, py

tkadlec:
Basic gist here for anyone following along: 

- We collect requests via an extension
- We also collect requests via underlying logs

We then loop through the extension requests to merge them with the requests from the logs, so that we can get more information. 

Since we never broke the loop, it was possible for the same extension request to be matched to multiple log requests.

scottjehl:
👍

PR #453, opened 2021-12-01
Fixed the User Agent override in Chrome to support Client Hints

pmeenan:
This is necessary for any page that requires client hints to be testable (otherwise it will error out with "too many redirects").


jefflembeck (approved):
Nice! Thanks!

PR #456, opened 2021-12-02
Fixed UA emulation

pmeenan:


pmeenan:
@jefflembeck @tkadlec @scottjehl FYI, I went ahead and merged this right away because it was fixing prod issues with mobile emulation.

tkadlec:
Thanks, @pmeenan

PR #457, opened 2021-12-07
Fixed Chrome CPU throttling

pmeenan:


tkadlec:
Adding https://bugs.chromium.org/p/chromium/issues/detail?id=1271559#c14 for reference.

tkadlec (approved):

tkadlec:
@pmeenan Given that a fix for this landed in Chrome Dev, what, if anything, can we pull back out here?

pmeenan:
@tkadlec Both of the command-line options added can be pulled back but not until Chrome 99 reaches stable (I think it is 99 that just branched).

PR #459, opened 2022-01-06
feat(wappalyzer): Add option to disable Wappalyzer

### LINKED ISSUE #449: Make Wappalyzer Optional
In a quick spot check of 10 URL's, the Wappalyzer check was one of the slowest parts of the WebPageTest agent. Let's add a flag to make Wappalyzer optional so that in situations where it's not being used, or speed of results is important, we can disable it.

tkadlec:
This adds an option (`wappalyzer`) that can be passed with a value of `0` to disable Wappalyzer detection on a given run.

It defaults to `1` to ensure that we align with what WebPageTest has defaulted to up to this point, and don't break things for folks.

Closes #449

### DIFF
--- a/internal/devtools_browser.py
+++ b/internal/devtools_browser.py
@@ -298,7 +298,8 @@ class DevtoolsBrowser(object):
             # Run the video post-processing
             if self.use_devtools_video and self.job['video']:
                 self.process_video()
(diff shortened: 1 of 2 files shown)


jefflembeck (approved):
This seems smart and good. 🐑  it

PR #468, opened 2022-02-08
Updated blink feature and css property mappings

pmeenan:
Looks like it got stale to the point where stable now has features not currently in the lookup.  Doesn't break anything for testing but is used by the web almanac.


tkadlec (approved):
Looks good to me!

PR #479, opened 2022-02-28
Move page_data calculations to the agent

pmeenan:
This ports all of the post-processing done to the page-level metrics in page_data.inc to the agent (and runs on each step of a given test independently).

This is a necessary step towards the agents being able to generate the JSON test result or HAR directly without relying on the server to run calculations on the raw data.

There will be a corresponding server-side change to not re-calculate all of the metrics when they were already calculated on the edge.

Eventually, the agent can delete the support files that are only used as part of the metric calculations but they are still uploaded with the test result for now, just to be safe.


pmeenan:
Argh, sorry, I had a bit of a branch nightmare and this got moved and now includes all of the changes along the way (and is what we are running in the HTTP Archive):
* Calculate the page stats in the agent
* Generate the HAR directly on the agent (if requested by command-line param or by a job option with a cloud storage destination)
* Updated Wappalyzer definitions
* Added support for pulling work from pubsub and posting results directly to cloud storage (without routing work through the WPT server)
    * The results uploaded to cloud storage are suitable for being restored to a WPT server for visualizing

tkadlec:
@pmeenan Heh...been there myself. No worries. Does look like there are some conflicts though that would stop me from being able to merge this in....

pmeenan:
@tkadlec Thanks, resolved the readme conflict so it should be good to go again.

PR #491, opened 2022-04-11
Add the sniffed image type and image metadata to the output

pmeenan:
Preparing for being able to extract more metadata from the original images which will be passed through to image_info.

Also improved the avif image sniffing to be more broad and also catch animated avif images.


tkadlec (approved):
This looks awesome! Thanks, @pmeenan!

PR #492, opened 2022-04-13
Fixing headers and other indicators for proxied resources

tkadlec:
This brings back headers, render blocking indicators, unused preloads, etc for URL's that get intercepted thanks to overrideHost


pmeenan:
LGTM.  Only nit is that `proxiedUrl` implies that the host override was for proxying. It's a common use case but not necessarily the only reason the host was substituted. Not sure if it is worth bikeshedding on the name but maybe overriddenUrl or rewrittenUrl

tkadlec:
Good call. I'll swap that.

PR #493, opened 2022-04-15
Added debug logs, Profiling, debug mode(with cmd Arguments), and new process_cpu_times() with 4x speed

sammeboy635:
Let me know if I'm doing something wrong or if I need to change something.
MAJOR ADDS:
* Added logs.write functions through out the code base.
* Removed some local imports that should be global imports
* internal/wptutil.py: This Utils file was added to help profile and log (singleton class)
* internal/support/devtools_parser.py: self.process_cpu_times_new() new function ready to go (Performance Benefit Below).
* internal/support/devtools_parser.py: self.extract_net_requests_new() work in progress (not implemented)

-Added a --debug: doesn't interfere with testspec.json debug mode.
Screen Shot 2022-04-15 at 8 27 34 AM

-Enabled through arguments /logging/log.txt ($python3 wptagent.py --debug)
Screen Shot 2022-04-14 at 9 58 01 PM

Profiler
Light Testrun ($python3 wptagent.py --debug --testurl light)
Screen Shot 2022-04-14 at 9 45 43 PM

Heavy Testrun ($python3 wptagent.py --debug --testurl heavy)
Screen Shot 2022-04-14 at 9 46 42 PM

### DIFF
--- a/internal/browsers.py
+++ b/internal/browsers.py
@@ -6,11 +6,14 @@
 """Main entry point for controlling browsers"""
 import logging
 import os
+from internal.wptutil import LogSingleton as logs
+
 
 
 class Browsers(object):
     """Controller for handling several browsers"""
     def __init__(self, options, browsers, adb, ios):
+        logs.write("INIT Browsers")
         try:
             import ujson as json
         except BaseException:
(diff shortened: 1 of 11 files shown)


tkadlec:
This is some exciting stuff, @sammeboy635! Thanks for doing this!

To make it easier to review and merge, could we close this PR and then file a few smaller ones?

If possible, I'm thinking:

- PR for adding logs.write functions 
- PR for removing some local imports that should be global imports
- PR for internal/wptutil.py (not sure if this needs to be paired with the logs.write PR or not)
- PR for self.process_cpu_times_new()

And then another PR for the other method you're working on when it's ready?

PR #496, opened 2022-04-19
Merging Performance Improvements

sammeboy635:
Committed two changes. Let me know if you need any more information.

### DIFF
--- a/internal/support/devtools_parser.py
+++ b/internal/support/devtools_parser.py
@@ -12,6 +12,8 @@ import os
 import re
 import sys
 import time
+from numpy import sum
+from math import ceil
 HAS_FUTURE = False
 try:
     from builtins import str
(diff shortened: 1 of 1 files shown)


internal/support/devtools_parser.py
```diff
@@ -877,6 +879,16 @@ def process_netlog_requests(self):
                    'tls_cipher_suite': 'tls_cipher_suite',
                    'uncompressed_bytes_in': 'objectSizeUncompressed'}
         if self.netlog_requests_file is not None and os.path.isfile(self.netlog_requests_file):
+            re1 = re.compile(r'^HTTP\/1[^\s]+ (\d+)').search
```
pmeenan, line 882: Just a drive-by comment, but any chance you can use descriptive names for the re's instead of 1-11? like re_http, re_status, re_content_type, etc?

sammeboy635: Hey, thanks for your feedback. That does seem to make more sense to have a better naming scheme for the regex. I'm still in the learning phase of working on a repo, so any feedback is greatly appreciated.

Issue #437, opened 2021-09-29
Provide a way to include certain binaries in the test results


rviscomi:
Over at HTTP Archive we would like to analyze how Wasm files are built for our annual Web Almanac "state of the web" report. The trouble is that response bodies from WPT are only included for text-based resources, excluding Wasm files.

It would be helpful for us if WPT provided a mechanism to specify which binaries' response bodies to include in the results, or to add a one-off exemption for Wasm.

Related issue: https://github.com/HTTPArchive/httparchive.org/issues/416

PR #497, opened 2022-04-19
Collect stats from wasm files

pmeenan:
This adds extraction of WASM stats from wasm binaries using a [fork](https://github.com/HTTPArchive/wasm-stats) of the tool used for last year's web almanac (modified to run on individual wasm files).

It adds a pre-built binary (and only works on) 64-bit linux x86 (pretty much all VM-based WPT agents). It is optimized for size as much as possible but is still adding a 400KB binary blob to the repo (we have binary blobs for tcpdump and tethering code so that's not that out of the ordinary). If there is interest, I can also add MacOS and Raspberry Pi binaries but I expect Linux will cover most needs.

Fixes https://github.com/WPO-Foundation/wptagent/issues/437
Fixes https://github.com/HTTPArchive/almanac.httparchive.org/issues/2914


tkadlec:
Looking solid! 

> If there is interest, I can also add MacOS and Raspberry Pi binaries 

How hard would it be to add the MacOS binary here?

pmeenan:
On it.  Installing Rust on my MacOS system now. Will update the PR shortly

pmeenan:
Done. Looks like the MacOS binary was a little bit smaller (385kb)

pmeenan:
@tkadlec ping.

Not urgent but this is the only patch we use in the HTTP Archive that is still separate.

PR #501, opened 2022-04-25
removed redundant testinfo start time set.

sammeboy635:
Running with (python3 wptagent.py -vvvv --location test --key 12546 --testurl 'https://google.com' --testspec testspec.json --testout url --xvfb)

job['testinfo']['started'] is caught on exception because 'testinfo' wasn't checked if it was a key first but earlier in the code was a job['testinfo']['started'] = job['started'] at line 579 with the proper check first. So I removed the redundant set.

No_check for testinfo

### DIFF
--- a/internal/webpagetest.py
+++ b/internal/webpagetest.py
@@ -607,7 +607,6 @@ class WebPageTest(object):
                     with open(job_path, 'wt') as f_out:
                         json.dump(job, f_out)
                     self.needs_zip.append({'path': job_path, 'name': 'job.json'})
-                    job['testinfo']['started'] = job['started']
                 # Add the non-serializable members
                 if self.health_check_server is not None:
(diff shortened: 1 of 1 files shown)


internal/desktop_browser.py
```diff
@@ -359,6 +359,8 @@ def stop(self, job, task):
 
     def wait_for_idle(self, wait_time = 30):
         """Wait for no more than 50% CPU utilization for 400ms"""
+        if self.options.noidle:
```
sammeboy635, line 362: I noticed that the --noidle flag was not checked inside the desktop_browser.
Here is the output from each run.
Before Fix: https://pastebin.com/KXT6QUtn
After Fix: https://pastebin.com/B8CK7Mgb
--noidleNotworking
after_noidle_fix

PR #506, opened 2022-05-09
Fixed two try except unused exception/ logging.Exception which should…

sammeboy635:
… be logging.exception
Minor fixes to reduce the amount of errors shown in Pylint. 
No function is called logging.Exception(), someone most likely meant logging.exception(). I removed the "Exception" as well because it is unused in the except.

### DIFF
--- a/internal/android_browser.py
+++ b/internal/android_browser.py
@@ -251,8 +251,8 @@ class AndroidBrowser(BaseBrowser):
                         logline = '{0:d},{1:d},-1,-1\n'.format(snapshot['time'], snapshot['bw'])
                         logging.debug(logline)
                         gzfile.write(logline)
-                except Exception:
-                    logging.Exception("Error processing usage queue")
+                except:
+                    logging.exception("Error processing usage queue")
                 gzfile.close()
         if self.tcpdump_enabled:
             tcpdump = os.path.join(task['dir'], task['prefix']) + '.cap'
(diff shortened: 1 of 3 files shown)


internal/android_browser.py
```diff
@@ -251,8 +251,8 @@ def on_stop_recording(self, task):
                         logline = '{0:d},{1:d},-1,-1\n'.format(snapshot['time'], snapshot['bw'])
                         logging.debug(logline)
                         gzfile.write(logline)
-                except Exception:
```
pmeenan, line 254: The `Exception` here is for identifying what kinds of Exceptions the handler wants to catch (all). The [docs](https://docs.python.org/3/tutorial/errors.html) don't really say what the behavior is if it isn't specified.

pmeenan (changes requested):

internal/android_browser.py
sammeboy635: Thanks for clarifying that!

internal/support/visualmetrics.py
```diff
@@ -901,9 +901,7 @@ def frames_match(image1, image2, fuzz_percent,
         if different_pixels <= max_differences:
             match = True
     else:
-        logging.debug(
-            'Unexpected compare result: out: "{0}", err: "{1}"'.format(
-                out, err))
+        logging.debug('Unexpected compare result: err: "{1}"'.format(err))
```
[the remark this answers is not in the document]
sammeboy635: Interesting I was using Pylint to check it and it doesn't seem to catch errors like that. You are correct thanks again for that catch.

pmeenan (approved):

tkadlec:
Looks great! Thanks for the fix @sammeboy635 and the review @pmeenan!

PR #507, opened 2022-05-09
Added #pylint: disable=import-error to the import that are giving pro…

sammeboy635:
…blems
Just trying to clean up Pylint a little more for Github Actions. Let me know if have any problems with this PR!

### DIFF
--- a/internal/process_test.py
+++ b/internal/process_test.py
@@ -21,13 +21,7 @@ if (sys.version_info >= (3, 0)):
     GZIP_READ_TEXT = 'rt'
     GZIP_TEXT = 'wt'
     string_types = str
-else:
-    from monotonic import monotonic
-    from urllib import quote_plus # pylint: disable=import-error,no-name-in-module
-    from urlparse import urlsplit # pylint: disable=import-error
-    GZIP_READ_TEXT = 'r'
-    GZIP_TEXT = 'w'
-    string_types = basestring
+
 try:
     import ujson as json
 except BaseException:
--- a/internal/firefox.py
+++ b/internal/firefox.py
@@ -149,8 +149,8 @@ class Firefox(DesktopBrowser):
         """Start Firefox using Marionette"""
         if self.must_exit:
             return
-        from marionette_driver.marionette import Marionette
-        from marionette_driver.addons import Addons
+        from marionette_driver.marionette import Marionette # pylint: disable=import-error
+        from marionette_driver.addons import Addons # pylint: disable=import-error
         args = ['-profile', '"{0}"'.format(task['profile']),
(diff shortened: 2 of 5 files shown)


internal/process_test.py
```diff
@@ -27,7 +27,7 @@
     from urlparse import urlsplit # pylint: disable=import-error
     GZIP_READ_TEXT = 'r'
     GZIP_TEXT = 'w'
-    string_types = basestring
+    string_types = basestring # pylint: disable=undefined-variable
```
sammeboy635, line 30: If you have any ideas on why this basestring is here please do let me know.

pmeenan: string_types is used on line 218 but the base type for a string is different in python 2 vs python 3.  We should probably make the call at some point to no longer support python 2.7 since it is EOL and has been for a while. AFAIK, all of the public agents (and all the private agents I know of) have been running on Python 3 for a while.

pmeenan: Now that we made the call to only support python 3, this should be cleaned up as part of the PR.

sammeboy635: If you would like I can cleanup all the files with the "if (sys.version_info >= (3, 0)):" removing the else portion of them or maybe remove the if statement entirely and just leaving the code. In a separate PR of course.

pmeenan: I'd recommend abandoning this PR and instead handle the import errors on a case-by-case basis:

- For Firefox, remove the marionette support which is 2.7-specific and flatten the classes so that the WebDriver implementation replaces the base Firefox implementation (and any 2.7 code-paths are removed).  That could be a single PR or a collection of them for simplifying the Firefox support.

- For Internet Explorer, remove the IE support since IE is officially dead.  If we decide to keep Edge support then that probably means combining the inheritence like the Firefox change but I'd probably rip out Edge support too since legacy Edge is also deprecated and Edge now uses the Chromium code path.  That is mostly just a matter of removing IE/Edge from the browser detection, the implementations and the .net ETW proxy process.  It will probably clean up a fair bit of code.

- For the platform-specific windows/mac imports, those make sense to disable to warnings as some devs may not have the modules installed.

- Then, after all that is done, go through and clean up the remaining places where python version is checked (except in wptagent.py for the compatibility check, make that require 3.x+).  Those can be separate PR's for each file or all together since it's basically the same code repeated.

sammeboy635: I'm really interested in trying to clean up this code base, so this sounds like a great plan moving forward. I'm little unfamiliar with the code for firefox and IE but let me see if I can get in the ballpark of what you would like for removing these implementation. Thanks

PR #508, opened 2022-05-09
Add support for $WPT_URL substitution into custom scripts

pmeenan:
For the HTTP Archive crawl custom metric, we need to make sure the links we extract are in the same origin as the page we originally navigated to (ignoring redirects). This allows for passing the WPT-navigated test URL through to custom metrics as an additional substitution (docs will be updated separately).

### DIFF
--- a/internal/devtools_browser.py
+++ b/internal/devtools_browser.py
@@ -571,14 +571,14 @@ class DevtoolsBrowser(object):
             accessibility_tree = None
             for name in self.job['customMetrics']:
                 custom_script = unicode(self.job['customMetrics'][name])
-                if custom_script.find('$WPT_URL') >= 0:
+                if custom_script.find('$WPT_TEST_URL') >= 0:
                     wpt_url = 'window.location.href'
                     if 'page_data' in self.task and 'URL' in self.task['page_data']:
                         wpt_url = '{}'.format(json.dumps(self.task['page_data']['URL']))
                     elif 'url' in self.job:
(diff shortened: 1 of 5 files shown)


internal/devtools_browser.py
```diff
@@ -571,6 +571,16 @@ def collect_browser_metrics(self, task):
             accessibility_tree = None
             for name in self.job['customMetrics']:
                 custom_script = unicode(self.job['customMetrics'][name])
+                if custom_script.find('$WPT_URL') >= 0:
```
tkadlec, line 574: Super nit here, maybe, but would $WPT_FINAL_URL or something like that make a bit more sense? @pmeenan

tkadlec (changes requested):

internal/devtools_browser.py
pmeenan: Maybe $WPT_NAVIGATED_URL but it's not necessarily the final one, it is what was entered before any redirects may have kicked in. In the most basic case, it is the URL that was passed to WPT with the test (only becomes something different if there is a script with multiple navigate commands).

tkadlec: Ah, gotcha. Yup, totally misread this last night.

Can we do $WPT_TEST_URL then?

pmeenan:
@tkadlec sure, done.

tkadlec (approved):

tkadlec:
Thanks, @pmeenan

PR #509, opened 2022-05-09
Add support for extracting image metadata with imagemagick

pmeenan:
This adds support for using imagemagick (already installed on all of the test machines) for extracting verbose image data. It compliments the data that comes out of exiftool and didn't add any significant overhead in my testing.

### DIFF
--- a/internal/optimization_checks.py
+++ b/internal/optimization_checks.py
@@ -989,6 +989,20 @@ class OptimizationChecks(object):
                                 check['info']['metadata'] = metadata
                         except Exception:
                             pass
+                        # Use imagemagick to convert metadata to json
+                        try:
+                            command = '{0} -verbose "{1}" json:-'.format(self.job['image_magick']['convert'], request['body'])
+                            subprocess.call(command, shell=True)
+                            magick = json.loads(subprocess.check_output(command, shell=True, encoding='UTF-8'))
+                            if magick is not None:
+                                if 'image' in magick:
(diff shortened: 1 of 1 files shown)


internal/optimization_checks.py
```diff
@@ -989,6 +989,20 @@ def check_images(self):
...
                         except Exception:
                             pass
+                        # Use imagemagick to convert metadata to json
+                        try:
+                            command = '{0} -verbose "{1}" json:-'.format(self.job['image_magick']['convert'], request['body'])
+                            subprocess.call(command, shell=True)
+                            magick = json.loads(subprocess.check_output(command, shell=True, encoding='UTF-8'))
+                            if magick is not None:
+                                if 'image' in magick:
+                                    remove = ['name', 'artifacts', 'colormap', 'proprties', 'version']
```
tkadlec, line 999: Should `proprties` be `properties`? Or is this like how the variable fonts folks decided to abbreviate `weight` to `wght`?

tkadlec (changes requested):

internal/optimization_checks.py
pmeenan: Argh.  Good catch, but looking at a few more samples, there can be some interesting details about the compressor used to create the image in there so I decided to leave it after all (well, remove the string which was just wrong, leave `properties`)

tkadlec (approved):

PR #511, opened 2022-05-10
import cgi.escape Deprecated since 3.2+ html.escape is 3.2+

sammeboy635:
@pmeenan  Let me know what you think of this switch.

Looking through Pylint errors I saw that cgi.escape didn't have a defined function called escape. Looking into it further, cgi.escape is deprecated since version 3.2 and html.escape takes its place 3.2+. 

https://github.com/PyCQA/bandit/issues/338

cgi escape()

https://stackoverflow.com/questions/20918146/how-to-do-html-escaping-in-python
cgi-dep

### DIFF
--- a/internal/message_server.py
+++ b/internal/message_server.py
@@ -72,9 +72,9 @@ class TornadoRequestHandler(tornado.web.RequestHandler):
             response += "</style>\n"
             response += "</head><body><div id='wptorange'></div>\n"
             if MESSAGE_SERVER.config is not None:
-                import cgi
+                import html
                 response += '<div id="wptagentConfig" style="display: none;">'
-                response += cgi.escape(json.dumps(MESSAGE_SERVER.config))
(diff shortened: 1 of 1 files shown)


internal/message_server.py
```diff
@@ -72,9 +72,9 @@ def get(self):
             response += "</style>\n"
             response += "</head><body><div id='wptorange'></div>\n"
             if MESSAGE_SERVER.config is not None:
-                import cgi
+                import html
```
pmeenan, line 75: You either need to special-case this for python >= 3.2 or we need to make the call the python 2.7 is no longer supported (in which case the change is fine).

My vote is to stop supporting 2.7 and clean up the code. @tkadlec any thoughts?

tkadlec: I agree. 3.2 has been out for 11 years now. :) Think it's safe to drop 2.7 support.

PR #513, opened 2022-05-12
Updated the chrome feature names mapping

pmeenan:
And updated the URLs to point to "main" instead of "master"


tkadlec (approved):
Good catch! Thanks.

PR #520, opened 2022-05-25
Fixed centos_install.sh, tested on centos 7 on google cloud

sammeboy635:
Let me know if you need anymore information or for me to test run a different command.
- switch to python3
- gcc-g++ and python3-devel required by google cloud libs
- switched pip install to look at the requirment.txt so we can have a single location for updating library requirements

Command used to test if its working.
centos7

What I deemed to be a successful run
centosResults

After Running ./centos_install.sh
done install

@pmeenan or @tkadlec It seems that centos installs a specific version of node JS. Should I update this to a newer version? or is this specific version required?

warning

### DIFF
--- a/centos_install.sh
+++ b/centos_install.sh
@@ -27,7 +27,7 @@ until python3 -m pip install --user -r .github/workflows/requirements.txt
 do
     sleep 1
 done
-curl --silent --location https://rpm.nodesource.com/setup_9.x | sudo bash -
+curl --silent --location https://rpm.nodesource.com/setup_16.x | sudo bash -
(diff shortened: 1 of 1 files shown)


pmeenan:
You should update the node install line:

```
curl --silent --location https://rpm.nodesource.com/setup_9.x | sudo bash -
```

to match what the [debian script](https://github.com/WPO-Foundation/wptagent-install/blob/master/debian.sh) does:

```
curl -sL https://deb.nodesource.com/setup_12.x | sudo -E bash -
```

(or better yet, update both of them to install Node 16 which is the current stable release)

PR #522, opened 2022-05-31
[review] Extract cmap, color, and OpenType features metadata from fonts.

bramstein:
For the HTTPArchive Almanac 2022 Fonts chapter we'd like to extract some additional metadata from fonts:

* The codepoints supported by a font (useful for detecting script support, subsetting overhead, detecting emoji/icon fonts, etc.)
* Which color formats (and versions) are supported, as well as color font palette support.
* Which OpenType features are supported and for which scripts and languages.

I've tested this pull request by running it against a random sample of about 8000 fonts and found no problems with it.

### DIFF
--- a/internal/font_metadata.py
+++ b/internal/font_metadata.py
@@ -199,7 +199,7 @@ def _read_features(ttf):
 
         return result
     except Exception:
-        loggin.exception('Error reading OpenType feature data')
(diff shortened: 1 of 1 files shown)


tkadlec:
This looks solid. Thanks @bramstein!

PR #538, opened 2022-07-05
Improve sorting of requests sent within the same millisecond

pmeenan:
The old trace-based netlog had sub-millisecond timing data so even requests sent all within the same millisecond would sort correctly (WPT keeps a start_time_float with the full start time for sorting pourposes).

The actual netlog has 1ms timing granularity on the event timestamps so the display of requests sent within the same millisecond can be random instead of the actual order the requests were sent.

WPT sorts the request by send time rather than request ID because prioritization and scheduling can cause browsers to send requests out of order from when they were discovered.

This patch adds the request ID (modulo 10000) as a nanosecond timestamp to the millisecond from the netlog to synthesize a float start time that will sort by request ID within the same millisecond while still not changing the reported start time.  There is an extreme edge case where the request ID's wrap over 10000 within a millisecond boundary but that is unlikely enough to happen that it's not a problem worth adding additional complexity to solve.

For example, [this](https://www.webpagetest.org/result/220701_AiDc2C_FVK/2/details/#waterfall_view_step1) test on production shows 2 of the fonts loading before the css but the preloads are at the end of the head and the actual request order is that they were sent after the CSS.  It's just that they were all emitted within the same millisecond.

Screen Shot 2022-07-05 at 10 23 24 AM

[Here](https://wpt.meenan.us/result/220705_NW_2/1/details/#waterfall_view_step1) is the same test on my dev server (different connection speed but works for demonstration) that has all of the head requests sent within the same millisecond but has them sorted in the correct order. The second font was actually sent second even though the response came back after the 3rd font.

Screen Shot 2022-07-05 at 10 25 16 AM

### DIFF
--- a/internal/support/netlog.py
+++ b/internal/support/netlog.py
@@ -416,7 +416,7 @@ class Netlog():
         # millisecond is still correct.
         for request in requests:
             if 'start' in request and 'netlog_id' in request:
-                request['start'] = float(request['start']) + (float(request['netlog_id'] % 1000) / 1000000.0)
+                request['start'] = float(request['start']) + (float(request['netlog_id'] % 10000) / 1000000.0)
         self.netlog_requests = requests
         return requests
 
--- a/internal/support/netlog.py
+++ b/internal/support/netlog.py
@@ -140,6 +140,7 @@ class Netlog():
         if 'url_request' in self.netlog:
             for request_id in self.netlog['url_request']:
                 request = self.netlog['url_request'][request_id]
+                request['netlog_id'] = request_id
                 request['fromNet'] = bool('start' in request)
                 if 'start' in request and request['start'] > last_time:
                     last_time = request['start']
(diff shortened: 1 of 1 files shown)


tkadlec (approved):
Clever. Thanks, @pmeenan!

paulirish:
(just curious…)

> The actual netlog …

this is `--log-net-log=file.txt` ?


> The old trace-based netlog …

is this the `netlog` category in tracing?

i haven't spent much time with it, but i would've assumed they'd do a similar thing to the JS Profile.. just take the real netlog data and wrap it up in trace events. but this suggests the instrumentation is slightly different, yeah?

pmeenan:
Correct but trace has it's own timestamps which are more granular. Problem
is that the trace-based netlog events don't always work, depending on the
process model and where the network service is (and broke with early
hints). The file-based netlog works always and can be streamed through a
pipe so it is more reliable.

PR #539, opened 2022-07-05
New way to create hostname-ipaddress

sammeboy635:
Working on fixing the new logger, I was trying to figure out how the hostname was generated in webpagetest.py and thought maybe we could clean this up and allow me to practice more on PRs). I tried a few different variants but this variant worked on all platforms(Windows, Linux, MacOS) getting the correct local IP address. Other variants worked on windows and Linux but macOS received a localhost IP, instead of a local IP address which is why this pc_name() method was chosen. 
 
As far as Line 108 - 121 in webpagetest.py, I can see its trying to get a mac address but a little unclear if it's important. I tested these lines in (Windows, Linux, MacOS, and a VM of Linux), and received no matches. If it is important maybe we could have a if/elif statement to make it run when needed and not every time.

@pmeenan please review and any feedback is very much appreciated! Thanks

### DIFF
--- a/internal/os_util.py
+++ b/internal/os_util.py
@@ -131,6 +131,10 @@ def get_file_version(filename):
     return version
 
 def pc_name():
-    with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as s:
-        s.connect(("8.8.8.8", 80))
-        return socket.gethostname() + "-" + s.getsockname()[0]
+    """ Grabs the hostname and Local IP address of the machine and returns hostname-IP """
+    try:
+        with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as s:
+            s.connect(("8.8.8.8", 80))
+            return socket.gethostname() + "-" + s.getsockname()[0]
+    except Exception as e:
+        logging.error("Error getting pc_name: ", e)
--- a/internal/os_util.py
+++ b/internal/os_util.py
@@ -138,3 +138,5 @@ def pc_name():
             return socket.gethostname() + "-" + s.getsockname()[0]
     except Exception as e:
         logging.error("Error getting pc_name: ", e)
+        
+    return platform.uname()[1]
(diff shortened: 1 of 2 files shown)


internal/os_util.py
```diff
@@ -128,3 +129,8 @@ def get_file_version(filename):
     except:
         logging.exception('Error getting file version for %s', filename)
     return version
+
+def pc_name():
+    with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as s:
```
pmeenan, line 134: This should be wrapped in a try-except with a default name returned if for some reason this throws.

pmeenan (changes requested):
Mostly looks good, just 2 minor tweaks.

The MAC address was a special case from when the public agents ran on VMWare instances where the image could be cloned and the name was auto-generated based on the MAC address, but only if it was in a special format from the VMWare OID.  That hasn't been used for years and is safe to rip out.

internal/os_util.py
```diff
@@ -128,3 +129,12 @@ def get_file_version(filename):
...
     return version
+
+def pc_name():
+    """ Grabs the hostname and Local IP address of the machine and returns hostname-IP """
+    try:
+        with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as s:
+            s.connect(("8.8.8.8", 80))
+            return socket.gethostname() + "-" + s.getsockname()[0]
+    except Exception as e:
+        logging.error("Error getting pc_name: ", e)
```
pmeenan, line 140: Almost there.  This should probably return platform.uname()[1] outside of the exception as a default so that self.pc_name always has a value like it did before.

sammeboy635: Got it thanks! if would like, please take a look at the other PR's.

pmeenan (approved):

pmeenan:
LGTM.

PR #540, opened 2022-07-05
Adding dir and file removing to os_util

sammeboy635:
Looking through the wptagent code base there are multiple locations that remove files and directory, which, 
- Have a lot of repeated code.
- Could allow us more control over keeping these files for debugging.
- Allows for a Unified logging.Debug result and with the new logging format we can see the function and line number that called it.

At some point I will create a PR to change these areas out with these os_util functions. If you agree, with it.


internal/os_util.py
```diff
@@ -128,3 +129,21 @@ def get_file_version(filename):
     except:
         logging.exception('Error getting file version for %s', filename)
     return version
+
+def util_remove_file(_file):
```
pmeenan, line 133: I'd get rid of the util_ prefix and make it just remove_file

internal/os_util.py
```diff
@@ -128,3 +129,21 @@ def get_file_version(filename):
...
+    """ Function to handle removing a single file"""
+    logging.debug("Removing File %s", _file, stacklevel=3)
+    try:
+        if os.path.isfile(_file):
+            os.remove(_file)
+    except Exception:
+        logging.exception('Error Trying to remove file = %s', _file)
+
+def util_remove_dir(_dir):
```
pmeenan, line 142: maybe remove_dir_tree() to make it clear that it is recursive?

pmeenan (changes requested):
Some minor tweaks to the names.

I assume there is going to be a follow-on PR that changes the existing call sites to use the new utility functions?

internal/os_util.py
sammeboy635: > 

I will be following up with another PR to change these locations out. Just wanted to get an opinion on it first.

pmeenan (approved):
LGTM.  Small nit, might be worth an optional param to not log on failure in case the caller is expecting that a file or directory may not exist.

tkadlec:
@sammeboy635 This looks great. Agree with @pmeenan on this:

> Small nit, might be worth an optional param to not log on failure in case the caller is expecting that a file or directory may not exist.

Can you make that change and then I'll merge.

sammeboy635:
> @sammeboy635 This looks great. Agree with @pmeenan on this:
> 
> > Small nit, might be worth an optional param to not log on failure in case the caller is expecting that a file or directory may not exist.
> 
> Can you make that change and then I'll merge.

I took out the logging.exception for now because this is a good point. Thanks

PR #541, opened 2022-07-05
added --noidleall to not conflict with --noidle

sammeboy635:
Needed for later commits IE Actions. 
- Allows for more consistent performance results
- Great for debugging so you don't have to wait the extra seconds.

### DIFF
--- a/internal/desktop_browser.py
+++ b/internal/desktop_browser.py
@@ -359,6 +359,8 @@ class DesktopBrowser(BaseBrowser):
 
     def wait_for_idle(self, wait_time = 30):
         """Wait for no more than 50% CPU utilization for 400ms"""
+        if self.options.noidle:
+            return
         import psutil
         logging.debug("Waiting for Idle...")
         cpu_count = psutil.cpu_count()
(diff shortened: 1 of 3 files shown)


internal/desktop_browser.py
```diff
@@ -359,6 +359,8 @@ def stop(self, job, task):
 
     def wait_for_idle(self, wait_time = 30):
         """Wait for no more than 50% CPU utilization for 400ms"""
+        if self.options.noidleall:
+            return
```
pmeenan, line 363: This one makes me a bit nervous.  The system-wide idle checks shouldn't have any side-effects but skipping the idle wait after launching the browser might end up trying to automate something before the browser is actually ready.

It's worth trying since --noidle is mostly meant for local dev testing but this one is likely to have side-effects that might also impact dev testing.

pmeenan (changes requested):
I wouldn't add a new flag.  Just expand the places where the existing --noidle applies.  It was meant to skip the idle checks.

internal/desktop_browser.py
sammeboy635: We could add a sleep with a const of x seconds, which would allow for consistency but also let the browser get ready.

pmeenan (approved):
LGTM

PR #542, opened 2022-07-07
Squashing support of 2.7 for firefox, Removed firefox_webdriver.py

sammeboy635:
Squashing firefox_webdriver.py into firefox.py. Copied all the functions from firefox_webdriver.py and replaced the corresponding functions in firefox.py. 
Ran a Test on Linux and everything ran fine. If you would like me to run anymore Test or see any output please do let me know.

@pmeenan please review!


internal/browsers.py
```diff
@@ -93,9 +93,6 @@ def get_browser(self, name, job):
             job['browser_info'] = self.browsers[name]
             if 'type' in self.browsers[name] and self.browsers[name]['type'] == 'Firefox':
                 if self.options.webdriver:
```
pmeenan, line 95: No need for the "webdriver" option anymore - it can be removed since it is always webdriver now.

pmeenan (changes requested):
Code looks good to me.  One small tweak is to get rid of the "webdriver" command-line option since it is no longer needed (also need to remove it from wptagent.py and the readme).

Make sure to test the agent with Firefox to make sure it actually still works with the code changes too (I haven't tried patching the change and testing locally).

pmeenan (approved):
Perfect, thanks. That cleans up the Firefox integration pretty well.

At some point we may want to create a WebdriverDesktopBrowser base class that holds most of the logic and just have the Firefox implementation handle the firefox-specific options.  That will make it easy to add any browser that supports webdriver but it can be a future refactor.  This looks good.

PR #543, opened 2022-07-12
internal.debug.trace module

sammeboy635:
Basic implementation of module Trace in location internal.debug. I Implemented trace to be like logging where trace would be called like (trace.debug(), trace.critical(), trace.info()). My reason to go with this design choice over creating a logging.trace() was as follows.
- My module would have to give logging.Trace() a definition or wptagent would fail to run.
- This separates functionality of the Global logging and the trace module since they are meant for different purposes.

I created a trace.md to give more information about it as well.

Following PR, will be including all the tracing points throughout the codebase.

@pmeenan please review

## Example Output
image


pmeenan:
I need to do some more research but this may take a little longer.  What exactly do you want to accomplish with the custom logger?

If I remember correctly:
- Add some new levels (trace being one of them)
- Colorize the output format

Also need to make sure it works with "debug" test mode for the agents which tees the log output to a file and sends it back with the results.

The built-in logger is extremely flexible and I am hoping it can be done with an output format filter and either a subclass with a close enough name to expose the aliases for logging.trace(), etc since it is only a utility function for passing through a log level to the underlying logger.

I haven't looked through the current implementation enough to fully understand what it is trying to do but it looks like it's a lot more complicated than I was hoping for.

pmeenan:
[This stack overflow answer](https://stackoverflow.com/questions/2183233/how-to-add-a-custom-loglevel-to-pythons-logging-facility/35804945#35804945) looks like it does exactly what I was looking for (aside from the output formatting).

sammeboy635:
The overall end goal for me was to get more information in Actions, so the reviewer has a easier time determining how beneficial a PR is. Some of the metrics I wanted in Actions are...
- Performance Metrics
- Data Integrity of final pagedata results
- More unit testing to confirm certain critical features still work.
- Testing on window, Linux, mac.

Which I'm interested to know if you would like to see these features added and how you would approach doing this if you do want these features? I didn't really get any feedback on how useful these features would be and thought it may be useful.

Trace is supposed to be a debugging performance monitoring tool, which will work in any .py you import trace into, which is why I think its a little better then the one in base_browser. Another benefit is since Python is locked to one core processing at a time (GIL) we can see when functions execute. For instance me as a new intern working on Wptagent was a bit of a challenge to understand what was going on. What helped me understand was being able to included trace points at any function, so I can see when a certain function is executed and at what point in the program it is executed.

Other things trace does
- Trace does add color to easily see error, which I thought might be a nice addition but completely not necessary.
- New format to the logging output

If you would like I could implement the stack overflow answer with logging.trace() but I would still need _MyFilter (selecting only the trace log level for main.log file) and _formatter(creating the record.elapsed_last call time) and then the util function to set definitions of logging.trace which you were talking about.

Or if you have a better solution that doesn't use logging, I wouldn't mind dropping this for your solution.

PR #549, opened 2022-07-28
internal.debug.trace module & logging.trace

sammeboy635:
Adding logging.trace in using internal.debug. Trace can be called from global logger and has all the attributes like any other attribute (debug, critical, info). The tracing points in the code will be added in a later PR but currently   -vvvvv does change the logging output to be more verbose in stdout. It doesn't affect any of the other logging settings (-vvvv and lower).

logging.trace is meant for determining timings throughout the code as well as be informative on when certain parts of the code execute.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@
 /*.cmd
 /alive.txt
 /10fps.sh
+/logging
 error.log
 loop.sh
 agent.sh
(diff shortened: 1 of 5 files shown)


mjkozicki:
Closing for another approach on logging. @sammeboy635 See https://github.com/catchpoint/WebPageTest.agent/pull/618/files

PR #556, opened 2022-07-28
Remove file

sammeboy635:
Replaced half of the current os.remove() functions with the new os_util.remove_file() function.  I also may have corrected a bug within wptagent.py removing the exit_file instead of the shutdown file. Please correct me if I'm wrong.

### DIFF
--- a/internal/chrome_desktop.py
+++ b/internal/chrome_desktop.py
@@ -15,6 +15,7 @@ import time
 from .desktop_browser import DesktopBrowser
 from .devtools_browser import DevtoolsBrowser
 from .support.netlog import Netlog
+from internal import os_util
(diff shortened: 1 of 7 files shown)


pmeenan:
FYI, I'm seeing an exception when I run with this patch:

```
Unhandled exception in test run: _log() got an unexpected keyword argument 'stacklevel'
```

It looks like stacklevel was added in Python 3.8 but "python3" on Ubuntu 18.04 is 3.6.  It looks like remove_file (and remove_dir_tree) should be updated to only use that param on > 3.8.  The logging call should probably be inside of a separate try/except block as well to make sure it doesn't add a failure mode.

PR #557, opened 2022-08-06
Use the streaming netlog as the primary source for response bodies

pmeenan:
This turns on response body byte capturing when the streaming netlogs are active and uses that as the primary source for response bodies (falling back to the legacy dev tools method if a body isn't available from the netlog). It also reduces the overhead since it is using memory-mapped pipes for the file I/O from netlog (on a background thread) instead of bouncing through the dev tools protocol.

This applies both to the response bodies captured all of the time for optimization checks as well as the response bodies that are stored when explicitly requested.

We were seeing some edge cases where service worker scripts could not retrieve the response bodies through dev tools and this completely resolves that issue.

### DIFF
--- a/internal/chrome_desktop.py
+++ b/internal/chrome_desktop.py
@@ -32,7 +32,6 @@ CHROME_COMMAND_LINE_OPTIONS = [
     '--disable-device-discovery-notifications',
     '--disable-domain-reliability',
     '--disable-background-timer-throttling',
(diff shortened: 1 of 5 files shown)


vibaldem:
Makes sense that getting events directly from netlog instead of getting it passed via dev tools is more reliable.

tkadlec (approved):

PR #560, opened 2022-08-19
Improved the logic for finding the base page request

pmeenan:
We were seeing a small number of pages in the HTTP Archive where the OCSP check was being flagged as the main document request.  This improves the logic pretty significantly by first preferring the "sec-fetch-dest: document" request header and then falling back to the content-type filtering (which now includes the mime type we were missing previously).


tkadlec (approved):
Nice! This looks great!

PR #561, opened 2022-08-19
Add reporting of the uncompressed size in the chunk timings

pmeenan:
This adds a "filtered" size entry to the chunk timings that carries the decompressed size of resources that are content-encoded (gzip, brotli). 

```json
    "chunks": [
        {
            "ts": 444,
            "bytes": 1387,
            "filtered": 2985
        },
        {
            "ts": 446,
            "bytes": 1396,
            "filtered": 3677
        },
        {
            "ts": 466,
            "bytes": 1396,
            "filtered": 3888
        },
```

This will make it possible to map individual chunks of HTML back to when they were flushed.

Some possible uses come to mind:
* Highlight the time in the waterfall when the LCP image element was delivered on the wire.
* Provide a UI that shows the raw HTML response body and match selection or cursor position to where in the waterfall entry that chunk of HTML was delivered.

### DIFF
--- a/internal/support/netlog.py
+++ b/internal/support/netlog.py
@@ -916,6 +916,8 @@ class Netlog():
             if 'has_raw_bytes' not in entry or not entry['has_raw_bytes']:
                 entry['bytes_in'] += params['byte_count']
                 entry['chunks'].append({'ts': event['time'], 'bytes': params['byte_count']})
+            elif entry['chunks']:
+                entry['chunks'][-1]['filtered'] = params['byte_count']
             if 'bytes' in params and self.on_response_bytes_received is not None:
                 try:
                     raw_bytes = base64.b64decode(params['bytes'])
--- a/internal/support/netlog.py
+++ b/internal/support/netlog.py
@@ -917,7 +917,7 @@ class Netlog():
                 entry['bytes_in'] += params['byte_count']
                 entry['chunks'].append({'ts': event['time'], 'bytes': params['byte_count']})
             elif entry['chunks']:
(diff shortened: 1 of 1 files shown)


internal/support/netlog.py
```diff
@@ -916,6 +916,8 @@ def process_url_request_event(self, event):
             if 'has_raw_bytes' not in entry or not entry['has_raw_bytes']:
                 entry['bytes_in'] += params['byte_count']
                 entry['chunks'].append({'ts': event['time'], 'bytes': params['byte_count']})
+            elif entry['chunks']:
+                entry['chunks'][-1]['filtered'] = params['byte_count']
```
tkadlec, line 920: _Super_ minor nit, but why 'filtered'? Would 'uncompressed' or something like that make more sense?

pmeenan: It is repeated in every chunk entry for every request so I was trying to keep it as short as possible (though I'm sure it will compress well since it is repeated).  I just picked "filtered" since it is part of "URL_REQUEST_JOB_FILTERED_BYTES_READ" but I'm happy to pick something more descriptive.

Happy to switch it to "uncompressed" if I was just being too cautious with the concern of the JSON size.

pmeenan: Maybe "inflated" or "decoded"?

pmeenan: Switched it to inflated - hopefully people won't infer that the request was using "deflate" compression but it's the same length as filtered and hopefully clearer.

tkadlec: Works for me!

PR #563, opened 2022-08-23
Adding our CDN isfcdn.com

mantico91:
Hi,
can you add our CDN?

Here you'll find all the information: https://www.ismartframe.com/en/isf-cdn/

Thank you in advance!


pmeenan:
Is .isfcdn.com the CDN CNAME?  I just did a dig lookup on the main www and it is pointing to .isf-edge.com

mjkozicki:
@mantico91 We can revisit once Pat's question is answered.

Issue #564, opened 2022-08-23
CWV element extraction script accidentally trips Chrome feature usage flags


pmeenan:
The [runtime script](https://github.com/WPO-Foundation/wptagent/blob/master/internal/support/chrome/inject.js) that is injected into pages to extract the element details from LCP, CLS, and paint events accidentally triggers Chrome's feature usage flags because it iterates over all of the element properties and accesses the values.

Instead of dumping the full element attributes, the script should be updated to just extract the bits that the WebPageTest UI and API need to make use of (URL, rectangle, background-image, innerHTML).

PR #565, opened 2022-08-23
Selectively extract element information for LCP and CLS events

pmeenan:
Fix https://github.com/WPO-Foundation/wptagent/issues/564

Avoids iterating over every property of the elements and tripping feature usage flags.

This should also reduce the overhead of the injected script slightly and extracts everything the UI needs to do the LCP and CLS analysis.

The element timing and paint timing observers were not being used yet so leave those out until they are actually added to the output.

### DIFF
--- a/internal/support/chrome/inject.js
+++ b/internal/support/chrome/inject.js
@@ -1,64 +1,65 @@
-var WptAgentFlatten = function(object) {
-    let contents = {}
-    for (const key in object) {
-        if (key == 'children') {
-            continue;
-        }
-        if (typeof(object[key]) === 'object') {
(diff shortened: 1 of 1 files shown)


stoyan:
looks good, I commented on your fork accidentally instead of here

pmeenan:
@stoyan fixed most of the nits - only one I wasn't 100% confident in was going to the dotted notation for the event attributes in case some of the attributes were missing sometimes.  Maybe that's not a valid concern or there's a better way to fail gracefully (if so, happy to make that change too).

stoyan (approved):

PR #569, opened 2022-09-20
Add optimization check data for all 2xx responses

pmeenan:
This applies the CDN check data to any response (that matches the request ID) and includes the optimization check results for any 2xx responses.

We are seeing some streaming cases where range requests (and 206 responses) are not getting the CDN attribution and other checks.


pmeenan:
FWIW, redirects still don't collect the CDN information and it may be worth doing a separate pass on the raw request data after it is processed to backfill that data.

PR #572, opened 2022-09-27
Backfill sequence numbers for any request missing them

pmeenan:
There was a previous fix for netlog requests that were missing sequence numbers and it shouldn't be possible for there to be any requests that don't have sequence numbers but there are some reports of some errors with $WPT_REQUESTS still so just to be absolutely positive, this fills in the sequence number for any request that may be missing one (allowing the sort to happen).

### DIFF
--- a/internal/devtools.py
+++ b/internal/devtools.py
@@ -823,8 +823,10 @@ class DevTools(object):
         if request_id in self.requests and 'fromNet' in self.requests[request_id] and self.requests[request_id]['fromNet']:
             events = self.requests[request_id]
             request = {'id': request_id}
-            if 'sequence' in events:
-                request['sequence'] = events['sequence']
+            if 'sequence' not in events:
+                self.request_sequence += 1
+                events['sequence'] = self.request_sequence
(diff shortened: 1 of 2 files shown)


stoyan (commented):
is there a case when you get, say `request['sequence']` 1,2,3, then nothing and you start adding "synthetic" 4,5,6 and then the 7th request comes with `request['sequence']` being 5?

pmeenan:
The sequence numbers added here are just for the time when they are extracted for custom metrics.  The underlying sequence numbers with the raw request data doesn't get modified.

It does take a bit of digging to see that get_requests calls get_request which constructs a new request object that it then populates with the request details so the request object here is pointing to a one-time list of one-time request objects so the sequence numbers won't be re-used.

That said, given the layers it takes to figure that out, it would be cleaner to just have get_request guarantee a sequence number is present and have it increment the main sequence numbers to be sure that there is no way it changes in the future.

PR update coming.

pmeenan:
@stoyan I moved the sequence updating to the raw underlying request data (and used the persistent sequence numbers).

Issue #573, opened 2022-09-29
Issue with HTTP/2 Redirect reporting


pmeenan:
WPT appears to be merging headers from the final redirect target with the original 302 response, leading to incorrect headers and response code information for some redirect requests.

Request #26 in [this test](https://www.webpagetest.org/video/compare.php?tests=220929_BiDcA4_CJ8-r:3-c:0) shows a redirect and claims the request URL is 
```
https://cdn-2.matterport.com/apifs/models/Crr8mopspuB/images/aru84H5ca4h/03.29.2021_09.06.24.jpg?t=2-087a9ff8e8ab7fd835fe3e8fdea7eda94b0ce38c-1664560232-1&k=apifs%2Fmodels%2FCrr8mopspuB%2Fimages%2Faru84H5ca4h%2F03.29.2021_09.06.24.jpg&width=400&disable=upscale&dpr=3
```
but the actual URL (from the netlog) was
```
https://my.matterport.com//api/v1/player/models/Crr8mopspuB/thumb?width=400&dpr=3&disable=upscale
```
that was redirected to the URL that was displayed (and is shown as request 36 in the waterfall).

The response headers are a mix of the original response headers from the 302 and the response headers from the redirected final response.

It could be coming from the dev tools or netlog processing but is likely to be from the dev tools processing since it is all treated as the same "request" and the splitting of the request ID stopped working.

Issue #26, opened 2017-07-06
Chrome video capture stops at 450 frames


pmeenan:
Looks like the devtools.screenshot method of recording is [hard-coded to a 450 frame limit](https://cs.chromium.org/chromium/src/content/browser/devtools/devtools_frame_trace_recorder.cc?type=cs&sq=package:chromium&l=31) which can be as short as 7.5 seconds at 60fps.

Video capture for Chrome should be moved to use the same methods as for Firefox (and be independent of Chrome itself).

pmeenan:
This is fixed on Linux where we switched to x11grab for all video capture.  Still need to fix for Windows and Mac

pmeenan:
Closing.  Moved to ffmpeg video capture on all platforms

PR #574, opened 2022-09-30
Fix redirect URL tracking from the netlog

pmeenan:
This fixes an issue where redirects were sometimes picking up incorrect data (URL and headers) from the final request instead of just the redirect response.

Not sure when/how this broke because it used to work for sure.

[Sample test](https://wpt.meenan.us/result/220930_RQ_1/1/details/#waterfall_view_step1) (See request #26 which is now correct)
Fixes https://github.com/WPO-Foundation/wptagent/issues/573


stoyan (approved):
looks good to me!

tkadlec (approved):

PR #575, opened 2022-10-05
Sanitize user-supplied command-line params

pmeenan:
In case something sneaks through the server-side sanitization, this sanitizes/quotes the command-line params when using the shell, particularly the browser ones that support user-provided params.


stoyan (approved):
Grazie!

PR #576, opened 2022-10-11
Fix extraction of background image URLs.

pmeenan:
Also:
* added support for other types of non-src images
* fixed edge-side processing of the LCP events
* Moved the edge-processing to before the debug log gets collected (so logs can be included)

For https://github.com/WPO-Foundation/webpagetest/issues/2455

### DIFF
--- a/internal/support/chrome/inject.js
+++ b/internal/support/chrome/inject.js
@@ -1,37 +1,40 @@
 new PerformanceObserver((entryList) => {
     for (const entry of entryList.getEntries()) {
-      try {
-          let event = {
-              name: entry.name,
-              entryType: entry.entryType,
-              startTime: entry['startTime'],
-              size: entry['size'],
-              url: entry['url'],
-              id: entry['id'],
-              loadTime: entry['loadTime'],
-              renderTime: entry['renderTime'],
-          };
-          if (entry['element']) {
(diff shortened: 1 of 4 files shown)


internal/support/chrome/inject.js
```diff
@@ -1,37 +1,40 @@
...
+                }
+                if (entry.element['currentSrc']) {
+                    event.element['currentSrc'] = entry.element.currentSrc;
+                }
+                try {
+                    let style = window.getComputedStyle(entry.element);
+                    if (style.backgroundImage && style.backgroundImage != 'none') {
+                        event.element['background-image'] = style.backgroundImage;
+                    }
+                    if (style.content && style.content != 'none') {
```
stoyan, line 31: Curious how do we use this? Because it can be a whole bunch of stuff other than `none`, I wonder why we filter out `none`

stoyan (approved):
looks good

internal/support/chrome/inject.js
pmeenan: 'none' is the default value that gets populated when it's not explicitly set. https://developer.mozilla.org/en-US/docs/Web/CSS/background-image#syntax

tkadlec (approved):

PR #580, opened 2022-11-09
Cleaned up Chrome command-line flags

pmeenan:
Removed some long-since deleted flags and added some new ones to match the Chrome launcher flags for disabling some chatty features.

cc @paulirish


internal/chrome_android.py
```diff
@@ -34,14 +34,19 @@
...
     '--net-log-capture-mode=IncludeSensitive',
-    '--load-media-router-component-extension=0',
     '--mute-audio',
     '--disable-hang-monitor',
     '--password-store=basic',
     '--disable-breakpad',
-    '--dont-require-litepage-redirect-infobar',
-    '--override-https-image-compression-infobar',
-    '--disable-fetching-hints-at-navigation-start'
+    '--disable-fetching-hints-at-navigation-start',
```
paulirish, line 41: I _think_ `--disable-features=OptimizationHints` may preclude this one. Looks kinda like it but I haven't verified in details.  No big deal either way.

internal/chrome_desktop.py
```diff
@@ -32,16 +32,19 @@
     '--disable-device-discovery-notifications',
```
paulirish, line 32: this flag is gone now and replaced (AFAIUI) by `--disable-features=OptimizationHints`

internal/chrome_desktop.py
```diff
@@ -68,11 +68,13 @@
     'Translate',
```
paulirish, line 68: you can drop `TranslateUI` (one line above). it was renamed.

paulirish (approved):
lgtm


* disable-site-isolation-trials is an interesting one.. ill dig into that a bit more. Depends on what kinda OOPIF behavior you want.. or are trying to emulate! Hmm.
* and yeah probably a good call on BackForwardCache, at least until either WPT or LH has a functional test for it.

btw.. for posterity.. this came from me taking another pass on https://github.com/GoogleChrome/chrome-launcher/blob/main/docs/chrome-flags-for-tools.md
and then creating [chrome-flags-for-tools comparison - Google Sheets](https://docs.google.com/spreadsheets/d/1n-vw_PCPS45jX3Jt9jQaAhFqBY6Ge1vWF_Pa0k7dCk4/edit#gid=1265672696) in the meantime.

pmeenan:
I was a bit worried on the timing of the Media Router flag but looks like it was [changed in 2020](https://chromium-review.googlesource.com/c/chromium/src/+/2330007) so there's no risk of bleed through in switching to the feature flag.

pmeenan:
Lists are sorted now and made the tweaks for more long-ago changes (after verifying none were more recent than 2020).

tkadlec (approved):
Yeah, this is looking great. Thanks for doing this, @pmeenan! And for your help too, @paulirish.

PR #582, opened 2022-11-28
Add Safari desktop support

pmeenan:
This adds support for testing Safari desktop without the need for an extension (was problematic to build, sign and install). It should "just work" once the Mac OS agents are updated and "Safari" is added to the list of browsers for the given location.

Sample result is [here](https://wpt.meenan.us/result/221128_ZiMS_4/1/details/#waterfall_view_step1).

The biggest thing that doesn't work (yet) is that the requests do not have URL's because the Safari logs (which are used to gather the request details) don't have URLs or headers but it does have all of the granular timing details so we can see a lot of things we couldn't see before (like wait time).

What DOES work:
* Video capture (and all video-based metrics)
* Page-level bytes and request counts
* Page-level onload time and fully loaded times
* Most script commands (except any that require accessing requests)
* Custom metrics
* CPU Utilization
* Bandwidth graph
* Request-level:
	* Mime types (colored waterfalls)
	* Response codes
	* Response sizes (compressed and uncompressed)
	* Prioritization details
	* Connection/socket
	* Connection set-up timings
	
Short-term future work will use resource timing from the page to try to backfill URLs for the requests and some other page-level timings.

When WebDriver bidi becomes available then we should get better access to the per-request details but that's probably still a ways out.


pmeenan:
Tried using resource timing to match request URLs but the data in resource timings is not unique enough to reliably match the requests from the logs so for now the requests will have to be without real URLs.

PR #584, opened 2022-11-30
Allow for running a script in an arbitrary frame context

pmeenan:
This enables cross-origin frame scripting for Chromium browsers (might work on iOS as well, haven't tested there yet).

It collects the available execution contexts from Chromium browsers and adds it to the page data json as execution_contexts:

```json
"execution_contexts": [
    {
        "id": 2,
        "origin": "http:\/\/127.0.0.1:8888",
        "name": ""
    },
    {
        "id": 3,
        "origin": "https:\/\/codepen.io",
        "name": ""
    },
    {
        "id": 4,
        "origin": "https:\/\/cdpn.io",
        "name": ""
    }
],
```

It also adds a new script command, `setExecutionContext` that lets you match one of the available contexts by id, origin or name (though name is always blank as best as I can tell) and any future exec commands will target that context (and it can be reset to the main document by not passing a match).

```
navigate	https://codepen.io/juno_okyo/pen/yOjaEZ
setExecutionContext	origin=https://cdpn.io
execAndWait	document.getElementById('name-input').value = 'Bob'; fetch('https://cdpn.io/blank.html');
```

### DIFF
--- a/internal/devtools.py
+++ b/internal/devtools.py
@@ -88,6 +88,8 @@ class DevTools(object):
         self.main_thread_blocked = False
         self.stylesheets = {}
         self.headers = {}
+        self.execution_contexts = {}
+        self.execution_context = None
         self.trace_parser = None
         self.prepare()
         self.html_body = False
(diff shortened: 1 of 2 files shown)


tkadlec (approved):
Looks awesome!!

manpreet-compro:
Hi @pmeenan 
Thanks for adding this option, highly useful. However I am not able to reset the execution context to main after performing some action inside iframe. Could you pls provide an example of that. Thanks!

PR #587, opened 2022-12-05
Stop disabling BF Cache for Chrome

pmeenan:
Originally there was some concern that the feature would cause problem for multi-step navigations since it keeps the old renderer around but it ends up to not be an issue: https://wpt.meenan.us/result/221205_ZiC4_2/

Leaving BF Cache enabled will make sure to test multi-step navigations like real visitors, including any background activity that isn't silenced immediately for more accurate results (it also allows for testing the improvements for back navigations).


jefflembeck (approved):
Sweet!

PR #589, opened 2022-12-06
Improve traffic-shaping compatibility

pmeenan:
Not a critical change since I don't believe any of the VM environment use the hardware acceleration offloads but discovered this one when I was working on the Ubuntu 22.04 support.  This turns off the TCP offload features that can cause packets to bypass traffic shaping for systems that support hardware offload.  It's a no-op for other systems but doesn't hurt.

Everything else in the agent worked fine out-of-the-box on 22.04 (technically this did as well)


pmeenan:
Tis is ready to merge for Ubuntu 22.04 support (currently live of the HTTP Archive crawl).

tkadlec (approved):

PR #596, opened 2023-01-09
Fix json encoding of testinfo for Ubuntu 22.04

pmeenan:
More likely than not, the behavior changed in a python revision between 3.4 and whatever is currently running on ubuntu 22.04 but the 'testinfo' dictionary entry containes a few live objects when it is cloned from the current job (TrafficShaper and MessageServer).  Previously it would just skip encoding those fields and continue.

This change makes it explicit and interprets and objects that can't be serialized to json as an empty string.

This fixes the agents running on Ubuntu 22.04 when the work they are doing is not generated from the WPT UI and they are synthesizing the test info (i.e. in the HTTP Archive or other batch modes).

Verified that this fixes the issue with using Ubuntu 22.04 for the HTTP Archive agents.

### DIFF
--- a/internal/webpagetest.py
+++ b/internal/webpagetest.py
@@ -1420,7 +1420,7 @@ class WebPageTest(object):
                     self.job['testinfo']['steps'] = max_steps
                 json_path = os.path.join(self.workdir, 'testinfo.json')
                 with open(json_path, 'wt') as f_out:
-                    json.dump(self.job['testinfo'], f_out)
+                    json.dump(self.job['testinfo'], f_out, default=lambda o: '')
                 self.needs_zip.append({'path': json_path, 'name': 'testinfo.json'})
 
             # Zip the files


stoyan:
Does this answer the question which python? :)

```
Run actions/setup-python@v3
Version 3.[6](https://github.com/WPO-Foundation/wptagent/actions/runs/3876394006/jobs/6610172792#step:3:7).8 was not found in the local cache
Error: Version 3.6.8 with arch x64 not found
```

pmeenan:
Not really.  That's whatever message is coming from the github action on the project that runs some tests in a test environment.  Looks like the github action needs to be updated and is having issues spinning up the test environment.

Looks like @sammeboy635 [set up the action](https://github.com/HTTPArchive/wptagent/commits/master/.github/workflows/wptagent_test.yml) and should probably update it.

sammeboy635:
It seems like ubuntu-latest doesn't have 3.6.8 supported. I can either set the ubuntu version to a lower version from this github page (https://github.com/actions/runner-images), or I can set the version of python to be a supported version, which will be listed (https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json). python-3.7.13 is lowest version supported in python3 in the latest ubuntu. Please let me know which way would be better. :)

pmeenan:
I'd recommend matching the prod environment.  Force 18.04 now and then update whenever the prod agents update.

Might also be worth trying a matrix that also includes newer OS's but at a minimum, a sanity-check to make sure it doesn't break prod would be best.

pmeenan:
FWIW, the discussion above is for #597 about the actions. This PR is clean and needed for Ubuntu 22.04 support (PR is already live in the HTTP Archive fork for the crawl).

tkadlec (approved):

PR #597, opened 2023-01-09
github actions fix, ubuntu-18.04 set

sammeboy635:
Issue with ubuntu-latest being updated and 3.6.8 python versionbeing no longer supported for github actions. changed ubuntu-latest to ubuntu-18.04.

Athough ubuntu-18.04 environment is now deprecated so we should consider updating.
https://github.com/actions/runner-images
https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json python version manifest

### DIFF
--- a/.github/workflows/wptagent_test.yml
+++ b/.github/workflows/wptagent_test.yml
@@ -15,7 +15,7 @@ permissions:
 jobs:
   build:
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
 
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


pmeenan:
There are a couple of pending PR's to fix some edge cases with 22.04 but otherwise it should be ready to go.

I'd recommend standing up 22.04 dev and QA instances first, make sure it works as expected and then migrate the prod agents to 22.04.  It should probably be done in the next couple of months before 18.04 reaches EOL.

Once that is done then it makes sense to change the action to use 22.04 explicitly.

tkadlec (approved):
Thanks @sammeboy635 !

PR #605, opened 2023-02-21
Use Chrome's new headless mode for lighthouse

pmeenan:
Chrome has a ["new" headless mode](https://www.selenium.dev/blog/2023/headless-is-going-away/#what-are-the-two-headless-modes) in 109+ (technically added in 96 but it's the official way now) that makes headless Chrome behave much more like the real Chrome with UI and uses all of the same code.  Of particular note for WebPageTest is that this makes the lighthouse audits (where headless is used) more accurate and fixes issues with the BF cache audit.

Before (for WebPageTest):
Screenshot 2023-02-21 at 10 49 24 AM

After (for WebPageTest):
Screenshot 2023-02-21 at 10 39 08 AM

Before (for Google Search):
Screenshot 2023-02-21 at 10 49 50 AM

After (for Google Search):
Screenshot 2023-02-21 at 10 38 36 AM

This also removes the hack of using the lighthouse legacy navigation flow (the issue in 10.0 that required it was fixed right before 10.0 was released).


mathiasbynens:
Note that the official launch for the new Headless mode (albeit still behind the `--headless=new` flag) is M112, which is currently the Canary channel. (We’ve been landing some fixes up until recently.)

Chrome 112 is scheduled for Beta between March 9–16, and Stable on March 29th.

pmeenan:
Not sure why the rebase isn't showing up cleanly over here but it is trying to pick up way more changes than the latest branch.  Going to close this and open a new PR now that Chrome 112 has rolled to stable.

PR #615, opened 2023-05-05
Update hosts.txt

stoyan:
copypasta from https://github.com/hoshsadiq/adblock-nocoin-list/blob/master/hosts.txt


pmeenan (approved):
LGTM

mjkozicki:
@pmeenan This is an older PR, is it still valid?

pmeenan:
Yep, looking at the source, that's still the latest update.  Miners abusing the system haven't been a problem recently so it's less urgent, particularly since the login requirements and rate-limiting but it's a good idea to keep ahead of them.

stoyan:
I never merged this because I have not tested it in a real set up. Looks safe but I was extra paranoid.

PR #618, opened 2023-05-24
feat: add logging by syslog

jefflembeck:
This change allows a user to opt into logging via syslog, in case they want the agent to open a socket out to a service that consumes logs.

WARNING - This likely does _not_ work on MacOS, since they changed the behavior of the syslog daemon

### DIFF
--- a/internal/logging_filter.py
+++ b/internal/logging_filter.py
@@ -1,0 +1,17 @@
+"""A filter for our logging handler"""
+
+import logging
+
+class LoggingFilter(logging.Filter):
+    def __init__(self, location, agent_name):
+        self.location = location
+        self.agent_name = agent_name
+        self.test_id = ''
+    def filter(self, record):
+        record.location = self.location
+        record.agent_name = self.agent_name
+        record.test_id = self.test_id
+        return True
+    def set_test_id(self, test_id):
+        self.test_id = test_id
+
--- a/internal/traffic_shaping.py
+++ b/internal/traffic_shaping.py
@@ -563,9 +563,20 @@ class NetEm(object):
                     subprocess.call(['sudo', 'tc', 'qdisc', 'add', 'dev', self.interface,
                                      'ingress'])
                     subprocess.call(['sudo', 'tc', 'filter', 'add', 'dev', self.interface, 'parent',
-                                     'ffff:', 'protocol', 'ip', 'u32', 'match', 'u32', '0', '0',
+                                     'ffff:', 'protocol', 'ip', 'prio', '2', 'u32', 'match', 'u32', '0', '0',
                                      'flowid', '1:1', 'action', 'mirred', 'egress', 'redirect',
                                      'dev', 'ifb0'])
+                    subprocess.call(['sudo', 'tc', 'filter', 'add', 'dev', self.interface, 'parent',
+                                     'ffff:', 'protocol', 'ip', 'prio', '1',
+                                     'u32', 'match', 'ip', 'protocol', '17', '0xff',
+                                     'match', 'ip', 'dport', '514', '0xffff',
+                                     'flowid', '1:1', 'action', 'pass'])
+                    subprocess.call(['sudo', 'tc', 'filter', 'add', 'dev', self.interface, 'parent',
+                                     'ffff:', 'protocol', 'ip', 'prio', '1',
+                                     'u32', 'match', 'ip', 'protocol', '6', '0xff',
+                                     'match', 'ip', 'dport', '22', '0xffff',
+                                     'match', 'ip', 'sport', '22', '0xffff',
+                                     'flowid', '1:1', 'action', 'pass'])
                 # Turn off tcp offload acceleration on the interfaces
                 try:
                     subprocess.call(['sudo', 'ethtool', '-K', self.interface, 'tso', 'off', 'gso', 'off', 'gro', 'off'])
--- a/internal/traffic_shaping.py
+++ b/internal/traffic_shaping.py
@@ -566,17 +566,13 @@ class NetEm(object):
                                      'ffff:', 'protocol', 'ip', 'prio', '2', 'u32', 'match', 'u32', '0', '0',
                                      'flowid', '1:1', 'action', 'mirred', 'egress', 'redirect',
                                      'dev', 'ifb0'])
-                    subprocess.call(['sudo', 'tc', 'filter', 'add', 'dev', self.interface, 'parent',
-                                     'ffff:', 'protocol', 'ip', 'prio', '1',
-                                     'u32', 'match', 'ip', 'protocol', '17', '0xff',
-                                     'match', 'ip', 'dport', '514', '0xffff',
-                                     'flowid', '1:1', 'action', 'pass'])
-                    subprocess.call(['sudo', 'tc', 'filter', 'add', 'dev', self.interface, 'parent',
-                                     'ffff:', 'protocol', 'ip', 'prio', '1',
-                                     'u32', 'match', 'ip', 'protocol', '6', '0xff',
-                                     'match', 'ip', 'dport', '22', '0xffff',
-                                     'match', 'ip', 'sport', '22', '0xffff',
-                                     'flowid', '1:1', 'action', 'pass'])
+                    for interface in [self.interface, 'ifb0']:
+                        for port in ['22', '514']:
+                            for direction in ['dport', 'sport']:
+                                subprocess.call(['sudo', 'tc', 'filter', 'add', 'dev', interface, 'parent',
+                                                 'ffff:', 'protocol', 'ip', 'prio', '1',
+                                                 'u32', 'match', 'ip', direction, port, '0xffff',
+                                                 'flowid', '1:1', 'action', 'pass'])
                 # Turn off tcp offload acceleration on the interfaces
                 try:
                     subprocess.call(['sudo', 'ethtool', '-K', self.interface, 'tso', 'off', 'gso', 'off', 'gro', 'off'])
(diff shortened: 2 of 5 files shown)


internal/webpagetest.py
```diff
@@ -749,6 +749,9 @@ def notify_test_started(self, job):
         """Tell the server that we have started the test. Used when the queueing isn't handled directly by the server responsible for a test"""
         if 'work_server' in job and 'Test ID' in job:
             try:
+                if 'saas_test_id' in job:
+                    logging.info("Test id: {0} started".format(job['saas_test_id']))
```
jefflembeck, line 753: I'd like to find a better way to format the saas_test_id in, but haven't yet.

pmeenan (changes requested):
Need to look at the [traffic shaping](https://github.com/WPO-Foundation/wptagent/blob/master/internal/traffic_shaping.py#L565) rules (netem for linux specifically) and make sure that the outbound rsyslog traffic is tagged in such a way that it doesn't go through the tc filters (and count against the test bandwidth when running a test)

wptagent.py
```diff
@@ -1225,6 +1231,17 @@ def main():
     logging.basicConfig(level=log_level, format="%(asctime)s.%(msecs)03d - %(message)s",
                         datefmt="%H:%M:%S")
 
+    if options.rsyslog and platform.system() == "Linux":
+        globals()['location'] = '' if options.location is None else options.location
+        globals()['agent_name'] = os_util.pc_name() if options.name is None else options.name
+        address = (options.rsyslog, logging.handlers.SYSLOG_UDP_PORT)
+        facility = logging.handlers.SysLogHandler.LOG_USER
+        syslog_handler = logging.handlers.SysLogHandler(address, facility)
```
pmeenan, line 1239: LOG_USER is the default, could leave facility off as it's an optional param.

internal/webpagetest.py
jefflembeck: so, I have

`globals()['test_id'] = job['Test ID'] if job['saas_test_id'] is None else job['saas_test_id']` going on, which allows for the the logging of the saas_test_id normally and then fallback to the Test ID otherwise. That _feels_ like it should handle this.

That being said, I do not believe there is a regular `Test ID`

internal/logging_filter.py
```diff
@@ -0,0 +1,10 @@
+"""A filter for our logging handler"""
+
+import logging
+
+class LoggingFilter(logging.Filter):
+    def filter(self, record):
+        record.location = '' if globals()['location'] is None else globals()['location']
+        record.agent_name = '' if globals()['agent_name'] is None else globals()['agent_name']
+        record.test_id = '' if globals()['test_id'] is None else globals()['test_id']
```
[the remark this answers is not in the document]
jefflembeck: @pmeenan I can just create an `__init__` and pass `location` and `agent_name` in, correct, right?

Something like

```python

class LoggingFilter(logging.Filter):
    def __init__(self, location, agent_name):
        self.location = location
        self.agent_name = agent_name
    def filter(self, record):
        test_id = globals()['test_id']
        record.location = self.location
        record.agent_name = self.agent_name
        record.test_id = '' if test_id is None else test_id 
        return True
```

and then when I call it:

```python
        location = '' if options.location is None else options.location
        agent_name = os_util.pc_name() if options.name is None else options.name
        f = LoggingFilter(location, agent_name)
        syslog_handler.addFilter(f)
```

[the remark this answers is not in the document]
jefflembeck: that'll construct the list of globals for each format, but it's not _quite_ as bad as doing it three times. Also, we know those values out the gates.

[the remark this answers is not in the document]
pmeenan: I think so.  Then we'd just need a global "current test ID" of some kind which should be much easier to carry around. You could just have the wptagent code pass a reference to the LoggingFilter instance into WptAgent() and WebPageTest() and have WebPageTest's code set a member property of the logging filter with the test_id as needed (and skip globals entirely).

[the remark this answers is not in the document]
jefflembeck: That's a whole lot of injection, but I'm into it.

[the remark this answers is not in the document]
jefflembeck: ok, I think I've got a solution

internal/webpagetest.py
```diff
@@ -78,6 +78,7 @@ def __init__(self, options, workdir):
         self.log_formatter = logging.Formatter(fmt="%(asctime)s.%(msecs)03d - %(message)s",
                                                datefmt="%H:%M:%S")
         self.log_handler = None
+        self.log_filter = None if options.log_filter is None else options.log_filter
```
jefflembeck, line 81:
```suggestion
        self.log_filter = options.log_filter
```

internal/webpagetest.py
```diff
@@ -488,6 +489,9 @@ def process_job_json(self, test_json):
         self.raw_job = dict(test_json)
         if job is not None:
             try:
+                if self.log_filter:
+                    self.log_filter.set_test_id(job['saas_test_id'])
+
```
jefflembeck, line 494:
```suggestion
                if self.log_filter and 'saas_test_id' in job:
                    self.log_filter.set_test_id(job['saas_test_id'])

```

jefflembeck:
> > Need to look at the [traffic shaping](https://github.com/WPO-Foundation/wptagent/blob/master/internal/traffic_shaping.py#L565) rules (netem for linux specifically) and make sure that the outbound rsyslog traffic is tagged in such a way that it doesn't go through the tc filters (and count against the test bandwidth when running a test)
> 
> Looks like you can add a rule to match the rsyslog traffic (and probably add ssh while you're in there) to `pass` and give the netem rule a lower priority. Example [here](https://unix.stackexchange.com/a/556221).

@pmeenan ok, had to go read about `tc` for a bit for this one, but I think I've got one in for syslog (give a pass for all traffic that has a destination port of 514 was the intention there, that oughta do it).

You think I should add one for port 22 as well?

jefflembeck:
@pmeenan I decided to go with `saas_test_id` _or_ `Test ID` if the former does not exist

internal/traffic_shaping.py
```diff
@@ -560,6 +560,16 @@ def install(self):
...
+                    subprocess.call(['sudo', 'tc', 'filter', 'add', 'dev', self.interface, 'parent',
+                                     'ffff:', 'protocol', 'ip', 'prio', '1',
+                                     'u32', 'match', 'ip', 'protocol', '17', '0xff',
+                                     'match', 'ip', 'dport', '514', '0xffff',
+                                     'action', 'pass'])
+                    subprocess.call(['sudo', 'tc', 'filter', 'add', 'dev', self.interface, 'parent',
+                                     'ffff:', 'protocol', 'ip', 'prio', '2',
+                                     'u32', 'match', 'ip', 'protocol', '6', '0xff',
+                                     'match', 'ip', 'dport', '22', '0xffff',
+                                     'action', 'pass'])
```
pmeenan, line 572: Can you test the netem settings on a test machine stand-alone with something like iperf to make sure the 514 UDP and 22 TCP traffic is not shaped while 53 UDP and 80/443 TCP are shaped?

I don't know how well the filters here for "pass" work with the qdisc's added later when shaping is enabled.

pmeenan: Also make sure the port 22 shaping is bypassed in both directions.

wptagent.py
```diff
@@ -1252,6 +1272,7 @@ def main():
     if options.collectversion and platform.system() == "Windows":
         get_browser_versions(browsers)
 
+    options.log_filter = log_filter
```
pmeenan, line 1275: I could be wrong, but I don't know if you can attach arbitrary properties to the options that comes back from argparse.  This is literally just the command-line options.  It is probably better to add another param to the WPTAgent (and WebPageTest) classes and pass the log filter directly.

pmeenan (commented):
Looks mostly good.  I'm uncomfortable enough with tc that I want to make sure you tested that it works as expected.

internal/traffic_shaping.py
jefflembeck: > I don't know how well the filters here for "pass" work with the qdisc's added later when shaping is enabled.

You are correct. These are out of order. Changing.

pmeenan: FWIW, it might be cleaner/easier to apply the filter to both interfaces, independent of protocol and both for src and dst (test traffic will never be from/to those ports).

Something like:

```python
for interface in [self.interface, 'ifb0']:
	for port in ['22', '514']:
		for direction in ['src', 'dst']:
			<tc filter command with direction, interface and port>
```

That will get you ~8 filter commands but it will make it easier to add ports later if needed and keep the code clean (once you figure out how to get the ordering/priorities working well).

jefflembeck: @pmeenan I need 514 to be `udp` only (protocol 17) so I'd need to do this in two separate loops, which is fine, but also is there much of a point at this time?

pmeenan: `need`?  Technically syslog can run over TCP even though the agent will only be configured for UDP. There will never be any test browser traffic on that port so bypassing unilaterally for TCP and UDP would be fine.

jefflembeck: oh, fair. didn't really think about that point. just need to make sure it's working with udp ... doesn't matter how else it works. ok.

jefflembeck: seems like

```python
                    for interface in [self.interface, 'ifb0']:
                        for port in ['22', '514']:
                            for direction in ['dport', 'sport']:
                                subprocess.call(['sudo', 'tc', 'filter', 'add', 'dev', interface, 'parent',
                                                 'ffff:', 'protocol', 'ip', 'prio', '1',
                                                 'u32', 'match', 'ip', direction, port, '0xffff',
                                                 'flowid', '1:1', 'action', 'pass'])
```
will get us there.

mjkozicki:
Closing as approach to logging will be default local. https://github.com/catchpoint/WebPageTest.agent/pull/638 @jefflembeck

PR #619, opened 2023-05-31
Flush DNS cache on ubuntu 22.04+

pmeenan:
Looks like there's yet another way.  See: https://askubuntu.com/questions/1404586/how-to-clear-dns-cache-in-22-04


stoyan (approved):
the more, the merrier

PR #624, opened 2023-06-17
Run pylint on files in the repo only in CI

szepeviktor:
`find` looks at the filesystem, `git ls-files` looks at git index.


szepeviktor:
Friendly ping 🏓

PR #638, opened 2023-09-06
wptagent log local logs syslog rfc5424

FedericoMulas:
ref 177487
* Added log handler for syslog rfc5424 to support structured logging, ported from https://github.com/jobec/rfc5424-logging-handler/tree/master with BSD license because seems frozen and had needed issues pending (only 2 files where needed).
* Added new `RotatedFileTransport(StreamTransport)` in `internal\rfc5424logging\transport.py` to support an external tool rotating the log file (e.g. logrotate, logstash, etc.).
* Added new file `rfc5424logging_context.py` and relative class `LoggingContext` to store structured logging context to enhance each log while keeping backward compatibility with other types of logs.
~~* Added additional option `--loglevel` and refactored log level option handling into `setup_logging` class to support current log format an the new introduced rfc5424 log format.~~
* Added additional option `--logformat` and refactored log level option handling into `setup_logging` class to support. It reuse `--verbose` option and `--log` to set the log file path and verbosity. Only supports  new introduced rfc5424 log format `syslog`.

After this PR is approved all log entries will be updated to add the new logging context.

### DIFF
--- a/internal/rfc5424logging/transport.py
+++ b/internal/rfc5424logging/transport.py
@@ -154,7 +154,7 @@ class UnixSocketTransport:
             socket_types = [self.socket_type]
 
         for type_ in socket_types:
-            # Syslog server may be unavailable during handler initialisation.
+            # Syslog server may be unavailable during handler initialization
             # So we ignore connection errors
             try:
                 self.socket = socket.socket(socket.AF_UNIX, type_)
@@ -214,7 +214,7 @@ class RotatedFileTransport(StreamTransport):
 
     def _start(self):
         """
-        Stat the transport opening the file and storing its stats.
+        Start opening the file and storing its stats as ID.
         """
         stream = open(self.file_name, "a+", encoding="utf-8")
 
--- a/internal/rfc5424logging/transport.py
+++ b/internal/rfc5424logging/transport.py
@@ -1,0 +1,202 @@
+import io
+import socket
+import ssl
+import sys
+
+if sys.version_info.major == 3:
+    text_stream_types = io.TextIOBase
+    bytes_stream_types = io.BufferedIOBase
+else:
+    text_stream_types = io.TextIOBase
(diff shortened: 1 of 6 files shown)


internal/rfc5424logging/transport.py
```diff
@@ -0,0 +1,248 @@
...
+    """Transport that allow an external process to rotate the file used as transport"""
+
+    def __init__(self, file_name):
+        self.file_name = file_name
+        self.identifier = None
+        StreamTransport.__init__(self, self._start())
+
+    def _start(self):
+        """
+        Stat the transport opening the file and storing its stats.
```
FedericoMulas, line 217:
```suggestion
        Start opening the file and storing its stats as ID.
```

internal/rfc5424logging/transport.py
```diff
@@ -0,0 +1,248 @@
...
+        self.open()
+
+    def open(self):
+        if self.socket_type is None:
+            socket_types = [socket.SOCK_DGRAM, socket.SOCK_STREAM]
+        else:
+            socket_types = [self.socket_type]
+
+        for type_ in socket_types:
+            # Syslog server may be unavailable during handler initialisation.
```
FedericoMulas, line 157:
```suggestion
            # Syslog server may be unavailable during handler initialization
```

mjkozicki:
@pmeenan If you could take a look, thanks.

mjkozicki (approved):

pmeenan (approved):
Mostly LGTM but the flag behavior is a bit confusing.  It might be cleaner to always use verbosity to set the log level and just use a different flag to switch logging into Rfc 5424 mode.  That way log file and verbosity params would be consistent. I'm good either way though.

FedericoMulas:
@pmeenan I refactored as suggested, works well too.

mjkozicki (approved):

PR #640, opened 2023-09-26
Logging fixes

pmeenan:
This changes the rfc5424 logging to only be imported when that mode of logging is being used and also adds an (optional) requirement for the `tzlocal` module which is required by the rfc5424 logging code.

This way the agents that are already deployed should install the module if it's not already on the systems.

Without these changes, the agents are likely to stop working as soon as they update unless tzlocal is installed on them (not part of the default install script).

### DIFF
--- a/wptagent.py
+++ b/wptagent.py
@@ -489,6 +489,7 @@ class WPTAgent(object):
 
         # Optional imports
         self.requires('fontTools', 'fonttools')
+        self.requires('pytz')
         self.requires('tzlocal')
 
         # Try patching ws4py with a faster lib
--- a/wptagent.py
+++ b/wptagent.py
@@ -8,7 +8,6 @@
 import atexit
 import logging
 import logging.handlers
-from internal.rfc5424logging import Rfc5424SysLogHandler, logging_context
 import os
 import platform
 import gzip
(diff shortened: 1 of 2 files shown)


wptagent.py
```diff
@@ -490,6 +489,7 @@ def startup(self, detected_browsers):
 
         # Optional imports
         self.requires('fontTools', 'fonttools')
+        self.requires('tzlocal')
```
FedericoMulas, line 492: https://github.com/catchpoint/WebPageTest.agent/pull/638/files#diff-292d079fe725eed06ead69e3675d638d8bda0c4b65aae36ebde665c7ed52a2beR20

it introduced a dependency on pytz too
```suggestion
        self.requires('tzlocal')
        self.requires('pytz')
```

FedericoMulas (changes requested):

pmeenan:
Thanks. Added `pytz` to the module installs and also updated the in-agent install script. It was a little less critical because it was pre-installed on the ubuntu images I was using but better safe than sorry.

FedericoMulas (approved):
LGTM!

mjkozicki (approved):
LGTM

pmeenan:
@mjkozicki could you merge when you get a chance (I'm not authorized for the agent - which I'm good with keeping that way)

PR #644, opened 2023-11-08
Improve stability of benchmark-based CPU throttle scaling

pmeenan:
This puts a limit on the maximum scale that will be applied to CPU throttling (defaults to 2) as well as records the benchmark results for the last 9 benchmarks (9 hours of agent running usually) and picks the median result so that benchmark outliers don't skew the CPU throttling.

### DIFF
--- a/internal/webpagetest.py
+++ b/internal/webpagetest.py
@@ -234,9 +234,31 @@ class WebPageTest(object):
                 hash_val.update(hash_data)
                 iteration += 1
             elapsed = monotonic() - start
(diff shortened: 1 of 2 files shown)


claud-io:
@pmeenan any reason for which this is closed? was not working as expected?

pmeenan:
It's working as expected and we have deployed it in the HTTP Archive.  I've had a few PR's that weren't getting any traction so I assumed it was no longer being actively maintained.  If that's not the case, I'm happy to restore it and reopen the PR.

PR #645, opened 2023-11-21
Use a fast 4G profile when running desktop lighthouse tests

pmeenan:


internal/traffic_shaping.py
```diff
@@ -98,13 +99,20 @@ def configure(self, job, task):
...
+                if self.options.android or ('mobile' in job and job['mobile']):
+                    # 1.6Mbps down, 750Kbps up, 150ms RTT
+                    rtt = 150
+                    in_bps = 1600000
+                    out_bps = 750000
+                else:
+                    # 10Mbps, 40ms RTT
+                    rtt = 40
+                    in_bps = 10240000
+                    out_bps = 10240000
```
lbartoli79, line 115: Should we do it even when `self.job['dtShaper']`is true? https://github.com/catchpoint/WebPageTest.agent/blob/4ac1576773d201c97af548723ae277432a172953/internal/devtools_browser.py#L856

pmeenan: Ahh, yes - good catch.  Update inbound.

internal/devtools_browser.py
```diff
@@ -854,7 +854,12 @@ def run_lighthouse_test(self, task):
             else:
                 cpu_throttle = '{:.3f}'.format(self.job['throttle_cpu']) if 'throttle_cpu' in self.job else '1'
                 if self.job['dtShaper']:
-                    command.extend(['--throttling-method', 'devtools', '--throttling.requestLatencyMs', '150', '--throttling.downloadThroughputKbps', '1600', '--throttling.uploadThroughputKbps', '768', '--throttling.cpuSlowdownMultiplier', cpu_throttle])
+                    if self.options.android or ('mobile' in self.job and self.job['mobile']):
+                        # 1.6Mbps down, 750Kbps up, 150ms RTT
+                        command.extend(['--throttling-method', 'devtools', '--throttling.requestLatencyMs', '150', '--throttling.downloadThroughputKbps', '1600', '--throttling.uploadThroughputKbps', '768', '--throttling.cpuSlowdownMultiplier', cpu_throttle])
+                    else:
+                        # 10Mbps, 40ms RTT
+                        command.extend(['--throttling-method', 'devtools', '--throttling.requestLatencyMs', '40', '--throttling.downloadThroughputKbps', '1024', '--throttling.uploadThroughputKbps', '1024', '--throttling.cpuSlowdownMultiplier', cpu_throttle])
```
lbartoli79, line 862: Is this a typo or is there a reason for this?
> `'--throttling.uploadThroughputKbps', '768'` (768 instead of 750)

Same question for this
> `'--throttling.downloadThroughputKbps', '1024', '--throttling.uploadThroughputKbps', '1024'` (1024 instead of 10240)

lbartoli79 (approved):

PR #663, opened 2024-04-18
Add modifiers support for keypress

MaryamHuntsperger:
Existing **kepress** command does not support modifiers to dispatch a key combination event such as **_Ctrl+A_**.

Using  Devtools Protocol (_https://chromedevtools.github.io/devtools-protocol/tot/Input/_) Input.dispatchKeyEvent modifiers, we have added a new value argument where the users can pass Alt, Ctrl or Shift and dispatch key combinations. The new format is:
**keypress** [tab] **key** [tab] **Alt|Ctrl|Shift**.
image

The new command is backward compatible and works as before if no value passed. In case of passing garbage values, it also ignores them. This is behavior is aligned with Devtools Protocol, when a key combination is at system level, it gets simply ignored.


jspurlin:
@claud-io / @lbartoli79 can you review this PR, it's been pending for 5 months 😮

Thanks!

claud-io:
@jspurlin sorry, will try to schedule this for the upcoming weeks

lbartoli79 (approved):

claud-io:
@jspurlin allows this change to work, we need to change also UI. We should be able to do it by next release

Issue #671, opened 2024-10-29
Update nodejs to 18.x for lighthouse compatibility


DmitriyStoyanov:
Right now in logs of agent we see next errors:
```
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'lighthouse@11.4.0',
npm WARN EBADENGINE   required: { node: '>=18.16' },
npm WARN EBADENGINE   current: { node: 'v16.20.2', npm: '8.19.4' }
npm WARN EBADENGINE }
```

So to fix it we need to update nodejs installation from 16.x to 18.x

lbartoli79:
Closed by #672

PR #672, opened 2024-10-29
Update Node.js to version 20.x

DmitriyStoyanov:
Update node.js to version 20.x in order to support existing version of lighthouse #671 that currently supported only 18.x or later versions


lbartoli79:
@DmitriyStoyanov we internally tested node 20, feel free to update the PR to use that version as it works fine and we should be good for some time

claud-io (approved):

lbartoli79 (approved):

PR #673, opened 2024-11-19
Fix mouseclick command check

lbartoli79:
`mouseClick` command check need to be lower case [webpagetest.py#L1003](
https://github.com/catchpoint/WebPageTest.agent/blob/aa58ff06757b7923c9d8ea3a56d1335e8dd8eef9/internal/webpagetest.py#L1003)


pmeenan (approved):
LGTM, thanks

claud-io (approved):

